PR #2, opened 2018-03-07
ELPP-3492 AMQP examples for PHP

giorgiosironi:
Tested on RabbitMQ.

Two alternatives, the second is preferred:

- `enqueue` is an abstraction layer but it cannot fully support all reliability requirements (publisher confirms rather than just the asynchronous basic publish)
- `amqplib` is the official PHP client for RabbitMQ, which implements AMQP 0.9.1 and publisher confirms.

### DIFF
--- a/events/php/composer.json
+++ b/events/php/composer.json
@@ -4,8 +4,8 @@
     "type": "project",
     "require": {
         "enqueue/enqueue": "^0.8.23",
-        "enqueue/amqp-lib": "*",
-        "php-amqplib/php-amqplib": "*"
+        "enqueue/amqp-lib": "^0.8.23",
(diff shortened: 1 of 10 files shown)


events/php/composer.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "elife/libero-events-php",
+    "description": "Demonstrations of events usage from a PHP application",
+    "type": "project",
+    "require": {
+        "enqueue/enqueue": "^0.8.23",
+        "enqueue/amqp-lib": "*",
+        "php-amqplib/php-amqplib": "*"
```
thewilkybarkid: `^2.0`?

thewilkybarkid (approved):
(Not actually tested it, but looks right.)

PR #3, opened 2018-03-07
ELPP-3492 AMPQ examples for Python

seanwiseman:
Tested on RabbitMQ, but compatible with any AMQP 0.9 implementation.

### DIFF
--- a/events/python/receive_kombu.py
+++ b/events/python/receive_kombu.py
@@ -1,0 +1,35 @@
+from kombu import Consumer, Exchange, Queue, Message
+
+from setup_kombu import (
+    ARTICLE_EXCHANGE_NAME,
+    DASHBOARD_QUEUE_NAME,
+    get_connection
+)
+
+
+def message_handler(body: str, message: Message) -> None:
+    print(f'[x] Received: {body}')
+    message.ack()
+
+
+if __name__ == '__main__':
+    with get_connection() as connection:
+
+        with connection.channel() as channel:
+
+            queue = Queue(name=DASHBOARD_QUEUE_NAME,
+                          exchange=Exchange(name=ARTICLE_EXCHANGE_NAME, type='fanout'))
+
+            consumer = Consumer(channel=channel,
+                                queues=queue,
+                                callbacks=[message_handler],
+                                no_ack=False)
+
+            print(' [*] Waiting for Messages. To exit press CTRL+C')
+            consumer.consume()
+
+            while True:
+                try:
+                    connection.drain_events()
+                except KeyboardInterrupt:
+                    break
--- a/events/python/receive_pika.py
+++ b/events/python/receive_pika.py
@@ -1,0 +1,20 @@
+from setup_pika import (
+    DASHBOARD_QUEUE_NAME,
+    get_channel
+)
+
+
+def message_handler(ch, method, properties, body):
+    print(f'[x] Received: {body}')
+
+
+if __name__ == '__main__':
+    with get_channel() as channel:
+        print(' [*] Waiting for Messages. To exit press CTRL+C')
+
+        channel.basic_consume(message_handler, queue=DASHBOARD_QUEUE_NAME, no_ack=True)
+
+        try:
+            channel.start_consuming()
+        except KeyboardInterrupt:
+            channel.stop_consuming()
--- a/events/python/receive_pika.py
+++ b/events/python/receive_pika.py
@@ -1,10 +1,13 @@
+import pika
+
 from setup_pika import (
     DASHBOARD_QUEUE_NAME,
     get_channel
 )
 
 
-def message_handler(ch, method, properties, body):
+def message_handler(channel: pika.channel.Channel, method: pika.spec.Basic.Deliver,
+                    properties: pika.spec.BasicProperties, body: str) -> None:
     print(f'[x] Received: {body}')
 
 
@@ -12,7 +15,7 @@ if __name__ == '__main__':
     with get_channel() as channel:
         print(' [*] Waiting for Messages. To exit press CTRL+C')
 
-        channel.basic_consume(message_handler, queue=DASHBOARD_QUEUE_NAME, no_ack=True)
+        channel.basic_consume(message_handler, queue=DASHBOARD_QUEUE_NAME, no_ack=False)
 
         try:
             channel.start_consuming()
--- a/events/python/receive_pika.py
+++ b/events/python/receive_pika.py
@@ -9,6 +9,7 @@ from setup_pika import (
 def message_handler(channel: pika.channel.Channel, method: pika.spec.Basic.Deliver,
                     properties: pika.spec.BasicProperties, body: str) -> None:
     print(f'[x] Received: {body}')
+    channel.basic_ack(method.delivery_tag)
 
 
 if __name__ == '__main__':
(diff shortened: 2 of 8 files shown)


events/python/receive_pika.py
```diff
@@ -0,0 +1,20 @@
...
+
+def message_handler(ch, method, properties, body):
+    print(f'[x] Received: {body}')
+
+
+if __name__ == '__main__':
+    with get_channel() as channel:
+        print(' [*] Waiting for Messages. To exit press CTRL+C')
+
+        channel.basic_consume(message_handler, queue=DASHBOARD_QUEUE_NAME, no_ack=True)
```
giorgiosironi: which part performs the consumer ack? Hidden in `start_consuming`?

events/python/send_pika.py
```diff
@@ -0,0 +1,37 @@
...
+        channel.exchange_declare(exchange=ARTICLE_EXCHANGE_NAME,
+                                 exchange_type='fanout',
+                                 durable=True)
+
+        # create queue, will skip if exists
+        channel.queue_declare(queue=DASHBOARD_QUEUE_NAME, durable=True)
+
+        # bind queue to exchange, will skip if already bound
+        channel.queue_bind(exchange=ARTICLE_EXCHANGE_NAME, queue=DASHBOARD_QUEUE_NAME)
+        channel.confirm_delivery()
```
giorgiosironi: this is about publisher confirms right?

events/python/send_pika.py
```diff
@@ -0,0 +1,37 @@
...
+if __name__ == '__main__':
+
+    with get_channel() as channel:
+        # create an exchange of type `fanout`, will skip if already exists
+        channel.exchange_declare(exchange=ARTICLE_EXCHANGE_NAME,
+                                 exchange_type='fanout',
+                                 durable=True)
+
+        # create queue, will skip if exists
+        channel.queue_declare(queue=DASHBOARD_QUEUE_NAME, durable=True)
```
giorgiosironi, line 22: durability :+1:

events/python/send_pika.py
```diff
@@ -0,0 +1,37 @@
...
+
+        while True:
+            msg = f'Test message: {time.time()}'
+            print(f'Publishing: {msg}')
+
+            # send a message
+            channel.basic_publish(exchange=ARTICLE_EXCHANGE_NAME,
+                                  routing_key=DASHBOARD_QUEUE_NAME,
+                                  body=msg,
+                                  properties=pika.BasicProperties(delivery_mode=DELIVERY_MODE_PERSISTENT))
```
giorgiosironi, line 37: persistence :+1:

events/python/send_pika.py
seanwiseman: Yes, will add a comment to make it clear.

events/python/receive_pika.py
seanwiseman: Hidden away, you can, however, pass `basic_consume` `no_ack=True` to disable the consumer ack if you wanted to. (its `False` by default)

events/python/receive_kombu.py
```diff
@@ -0,0 +1,35 @@
...
+from setup_kombu import (
+    ARTICLE_EXCHANGE_NAME,
+    DASHBOARD_QUEUE_NAME,
+    get_connection
+)
+
+
+def message_handler(body: str, message: Message) -> None:
+    print(f'[x] Received: {body}')
+    message.ack()
```
giorgiosironi, line 12: is this necessary given the `no_ack=False` later on?

events/python/receive_kombu.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+if __name__ == '__main__':
+    with get_connection() as connection:
+
+        with connection.channel() as channel:
+
+            queue = Queue(name=DASHBOARD_QUEUE_NAME,
+                          exchange=Exchange(name=ARTICLE_EXCHANGE_NAME, type='fanout'))
```
giorgiosironi, line 21: I wouldn't expect to have to know anything about the `Exchange` from a consumer point of view, only the queue name :thinking:

events/python/receive_pika.py
```diff
@@ -0,0 +1,24 @@
...
+from setup_pika import (
+    DASHBOARD_QUEUE_NAME,
+    get_channel
+)
+
+
+def message_handler(channel: pika.channel.Channel, method: pika.spec.Basic.Deliver,
+                    properties: pika.spec.BasicProperties, body: str) -> None:
+    print(f'[x] Received: {body}')
+    channel.basic_ack(method.delivery_tag)
```
giorgiosironi, line 12: same as for the other example, `no_ack=False` makes this unnecessary or is the other way around?

giorgiosironi: (could mean either "the library should not ack because the handler will do it" or "the handler should not have to ack")

events/python/receive_pika.py
```diff
@@ -0,0 +1,24 @@
...
+if __name__ == '__main__':
+    with get_channel() as channel:
+        print(' [*] Waiting for Messages. To exit press CTRL+C')
+
+        channel.basic_consume(message_handler, queue=DASHBOARD_QUEUE_NAME, no_ack=False)
+
+        try:
+            channel.start_consuming()
+        except KeyboardInterrupt:
+            channel.stop_consuming()
```
giorgiosironi, line 24: would be interesting to see how this can deal with a `SIGTERM`. The signal handler should call `stop_consuming`?

events/python/send_kombu.py
```diff
@@ -0,0 +1,47 @@
...
+            producer = Producer(channel)
+
+            while True:
+                msg = f'Test message: {time.time()}'
+                print(f'Publishing: {msg}')
+
+                producer.publish(
+                    body=msg,
+                    delivery_mode=DELIVERY_MODE_PERSISTENT,
+                    routing_key=DASHBOARD_QUEUE_NAME,
```
giorgiosironi: afaik a routing key should be optional?

events/python/receive_pika.py
seanwiseman: I found through later testing that if I did not manually `ack` in the handler then the messages were labelled as received but `unacked` by the server.

events/python/send_kombu.py
seanwiseman: You are right, added it as was following the docs, but have now removed it.

events/python/receive_pika.py
giorgiosironi: Good to know

events/python/receive_kombu.py
seanwiseman: That was what I expected. Though if you remove the exchange arg, you receive no messages 🤔 . Would require a further dip into the docs.

events/python/receive_pika.py
seanwiseman: Yes exactly. I just used `KeyboardInterrupt` because it was probably going to be how people stop this demo script. Getting the signal handler to call `stop_consuming` would be the way.

events/python/receive_kombu.py
seanwiseman: > I found through later testing that if I did not manually ack in the handler then the messages were labelled as received but unacked by the server.

Same deal here in practice.

events/python/receive_kombu.py
seanwiseman: I have seen in several places in the docs it wants consumers to define the `exchange` as well as the `queue` you want to consume from. Possibly a side effect of them supporting multiple transports? 🤔 whatever the reason I don't like it. Another way is to use some of there mixin classes but that looks overly verbose for what is needed.

giorgiosironi (approved):

PR #4, opened 2018-03-08
RFC: event bus

giorgiosironi:
I'm trying to use ADRs as a collaborative model for RFC (editable until it is inside a pull request). All history should be kept by the Github branch and the snapshot of the final version of an RFC would become the ADR.


adr/0001-events-system.md
```diff
@@ -0,0 +1,41 @@
...
+
+- durable exchanges and queues (AMQP 0.9.1)
+- persistent messages (AMQP 0.9.1)
+- consumer acknowledgements (AMQP 0.9.1)
+- producer confirms (RabbitMQ extension to the protocol)
+
+Use the following libraries for the supported programming languages:
+
+- PHP: [amqplib](https://github.com/php-amqplib/php-amqplib)
+- Python: ?
```
giorgiosironi: @seanwiseman to fill in

thewilkybarkid (approved):

adr/0001-events-system.md
```diff
@@ -0,0 +1,41 @@
...
+
+- durable exchanges and queues (AMQP 0.9.1)
+- persistent messages (AMQP 0.9.1)
+- consumer acknowledgements (AMQP 0.9.1)
+- producer confirms (RabbitMQ extension to the protocol)
+
+Use the following libraries for the supported programming languages:
+
+- PHP: [amqplib](https://github.com/php-amqplib/php-amqplib)
+- Python: [pika](https://github.com/pika/pika)
```
thewilkybarkid, line 33: Don't think it needs a wording change here, but we should probably use adapters (to allow for null/in-memory implementations for testing), which means these aren't hard dependencies.

giorgiosironi: Ok, we can put that in ADRs for services testing. Edit: though they should be provided by the SDK themselves.

PR #6, opened 2018-03-16
Add modular RELAX NG example

thewilkybarkid:
Contains an experiment to use RELAX NG as a schema for the data model/API. I've not used it before, but seems to solve some of the problems encountered with XSDs.

Modularity in XSDs is limited, and often results in open schemas, so the base Libero things would be defined, but they then allow any other elements etc, which remain undefined. This allows you to include your own things, but isn't really controlled. (Can be partially solved with abstract elements.)

RELAX NG allows you to decorate definitions, which makes extensibility straightforward. It also don't have the element-ordering constraint that XSD has.

This has some base Libero types (an article, some core definitions and then some extensions, eg DOIs), and an eLife article type that uses them and adds its own, then a demo article making use of it.


api/xml/libero/core/inline.rng
```diff
@@ -0,0 +1,27 @@
...
+    <define name="libero.text.class">
+        <choice>
+            <ref name="libero.text.i"/>
+        </choice>
+    </define>
+
+    <define name="libero.text.model">
+        <oneOrMore>
+            <choice>
+                <text/>
```
thewilkybarkid: Can't see a way to make sure that elements aren't empty/blank.

api/xml/libero/extensions/doi.rng
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<grammar xmlns="http://relaxng.org/ns/structure/1.0" ns="http://libero.pub">
+
+    <define name="libero.article.content" combine="interleave">
+        <optional>
```
thewilkybarkid: Can't see a way to make these extensions either optional or required, without having an extension for each (probably using a shared definition for the actual elements/attributes).

api/xml/elife/article.rng
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+
+<grammar xmlns="http://relaxng.org/ns/structure/1.0" ns="http://elifesciences.org">
```
giorgiosironi: "next generation" in a name doesn't age particularly well

api/xml/libero/core/inline.rng
```diff
@@ -0,0 +1,27 @@
...
+<grammar xmlns="http://relaxng.org/ns/structure/1.0"
+    datatypeLibrary="http://www.w3.org/2001/XMLSchema-datatypes">
+
+    <define name="libero.text.i">
+        <element name="i">
+            <ref name="libero.text.model"/>
+        </element>
+    </define>
+
+    <define name="libero.text.class">
```
giorgiosironi: is `class` common terminology from `relaxng` schemas?

thewilkybarkid: Not sure. I borrow these ideas from the XHTML 2.0 (ahem) schema - https://www.w3.org/TR/xhtml2/relax_module_defs.html.

api/xml/elife/articles/00569.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" ?>
+<?xml-model href="../article.rng"?>
```
giorgiosironi: If this was returned by the API the `xml-model` reference should go a URL, which is in the specification.

api/xml/elife/articles/00569.xml
```diff
@@ -0,0 +1,53 @@
...
+
+<article xmlns="http://libero.pub" xmlns:elife="http://elifesciences.org" xml:lang="en">
+
+    <id>00569</id>
+    <doi>10.7554/eLife.00569</doi>
+    <title>The genome sequence of the colonial chordate, <elife:genome>Botryllus
```
giorgiosironi: This `elife:` prefix and corresponding namespace seems to be the standard way for in-document extensions.

giorgiosironi: It's not clear yet how this document can be produced. Several cases:
- standard mandatory fields like `<title>`: the article store can easily store them.
- standard optional fields: same.
- extensions to the standard fields/new fields: the article store can expand what it accepts, but it needs to know about the full article schema. So if I want a IIIF service, it must get its schema extension into the article store somehow.
- other responses like `search`: sort them out separately, but they need to know the full schema too.

In a brute force way this would mean adding IIIF would mean redeploying the article service and potentially any dependants like `search`, creating a mutual dependency or a cascade of dependencies.

The possible options:
- having the article store use a standard article schema and have a `mashup` service add extensions from other services. This is unfeasible as it's too difficult to interpolate `<elife:genome>` in the right place.
- promote the concept of schema as data that is persisted rather than configuration. The article store should keep it somewhere in some kind of database, and use it to validate incoming data. Downstream services like `search` refer to the article store for validation.
- promote the concept of schema even more as a `schema` service. This service would then be able to offer validation to whoever needs it. This implies something we may do anyway, which is versioning of schemas: whenever a new extension like IIIF is introduced, the article schema version number is incremented and the article store and dependants have time to upgrade to deal with it. It is unfeasible to change the schema atomically with the content anyway (think of the hours or even just minutes to regenerate all the affected articles).

giorgiosironi: See Google docs for more ideas on service boundaries

thewilkybarkid:
This has been updated to be schema files for the API Dummy.

api-dummy/data/articles/0065-1737-azm-31-03-0367/1/en/front.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" ?>
+<?xml-model href="http://localhost:8081/schemas/scielo/article/front.rng"?>
```
giorgiosironi, line 2: big assumption, would something like `/schemas/...` work?

giorgiosironi: Perhaps it can be interpolated by the api-dummy when serving the file, with a default of `http://localhost:8081`

giorgiosironi (commented):
The grammars seem in the right place, unclear how to insert absolute URLs

api-dummy/data/articles/0065-1737-azm-31-03-0367/1/en/front.xml
thewilkybarkid: Stops Oxygen from being able to validate the files locally. Going to see if I can set up a mapping there somehow.

thewilkybarkid: Hmm, this is the mapping that it then overwrites. Probably something we could ignore for now, but need to think about. (Part of a bigger question, where should schemas live?)

thewilkybarkid:
:shipit:?

seanwiseman (approved):

giorgiosironi (approved):

giorgiosironi:
> Stops Oxygen from being able to validate the files locally

We'll soon need a portable alternative for validation

PR #7, opened 2018-03-27
Add API Dummy

thewilkybarkid:
Using data from #6.

### DIFF
--- a/api-dummy/.docker/nginx-server.conf
+++ b/api-dummy/.docker/nginx-server.conf
@@ -21,7 +21,4 @@ server {
     location ~ \.php$ {
         return 404;
     }
-
-    error_log /var/log/nginx/app_error.log;
-    access_log /var/log/nginx/app_access.log;
 }
--- a/api-dummy/.docker/nginx-server.conf
+++ b/api-dummy/.docker/nginx-server.conf
@@ -1,0 +1,27 @@
+server {
+    server_name localhost;
+    listen 80;
+    root /srv/app/public;
+
+    location / {
+        try_files $uri /index.php$is_args$args;
+    }
+
+    location ~ ^/index\.php(/|$) {
+        fastcgi_pass api-dummy-fpm:9000;
+        fastcgi_split_path_info ^(.+\.php)(/.*)$;
+        include fastcgi_params;
+
+        fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;
+        fastcgi_param DOCUMENT_ROOT $realpath_root;
+
+        internal;
+    }
+
+    location ~ \.php$ {
+        return 404;
+    }
+
+    error_log /var/log/nginx/app_error.log;
+    access_log /var/log/nginx/app_access.log;
+}
--- a/api-dummy/Dockerfile
+++ b/api-dummy/Dockerfile
@@ -1,0 +1,33 @@
+FROM php:7.1-fpm
+
+RUN mkdir -p /var/www && \
+    chown www-data:www-data /var/www
+
+RUN apt-get update && apt-get install -y \
+    git-core \
+    zip
+
+RUN docker-php-ext-configure opcache --enable-opcache \
+    && docker-php-ext-install opcache
+
+COPY .docker/scripts /root/scripts
+RUN /root/scripts/install-composer.sh
+
+RUN mkdir -p /srv/app && \
+    chown www-data:www-data /srv/app
+
+USER www-data
+WORKDIR /srv/app
+
+COPY --chown=www-data:www-data composer.json composer.lock ./
+RUN composer --no-interaction install --no-autoloader --no-scripts --no-suggest
+
+COPY --chown=www-data:www-data . ./
+
+ARG APP_ENV
+ARG APP_SECRET
+ENV APP_ENV="$APP_ENV"
+ENV APP_SECRET="$APP_SECRET"
+
+RUN composer dump-autoload
+RUN composer run-script post-install-cmd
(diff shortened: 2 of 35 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,20 @@
+version: '3'
+
+services:
+    api-dummy-fpm:
```
giorgiosironi, line 4: our current api-dummy ended up using the built-in web server because of the ease of use (only one image to specify, no configuration); an nginx and fpm setup is more realistic, but a dummy is not necessarily realistic?

thewilkybarkid: The skeleton doesn't need caching etc, but multiple requests would become synchronous, making the Journal appear slow.

It's not that much extra to add here too.

giorgiosironi: Skeleton-wise I think it's very good to have both nginx and fpm, since it's supposed to be touching roughly all pieces

api-dummy/.docker/nginx-server.conf
```diff
@@ -0,0 +1,27 @@
...
+        fastcgi_param DOCUMENT_ROOT $realpath_root;
+
+        internal;
+    }
+
+    location ~ \.php$ {
+        return 404;
+    }
+
+    error_log /var/log/nginx/app_error.log;
```
giorgiosironi: in the default image logs probably go to stderr, which makes `docker logs` collect them; may be unnecessary to override these two

api-dummy/Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM php:7.1-fpm
```
giorgiosironi, line 1: We can't really port the elife base images out here as they contain specific software like New Relic support

api-dummy/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN mkdir -p /var/www && \
+    chown www-data:www-data /var/www
```
giorgiosironi, line 4: a user to own the files making them not writable from `www-data` would be good, but can be easily postponed

api-dummy/bin/console
```diff
@@ -0,0 +1,27 @@
...
+set_time_limit(0);
+
+require __DIR__.'/../vendor/autoload.php';
+
+$input = new ArgvInput();
+$env = $input->getParameterOption(['--env', '-e'], $_SERVER['APP_ENV'] ?? 'dev', true);
+$debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env)) && !$input->hasParameterOption('--no-debug', true);
+
+if ($debug) {
+    umask(0000);
```
giorgiosironi: why is this necessary? All files and directories should be created by a single user at the moment

api-dummy/composer.json
```diff
@@ -0,0 +1,48 @@
...
+        "symfony/console": "^4.0",
+        "symfony/dependency-injection": "^4.0",
+        "symfony/flex": "^1.0",
+        "symfony/framework-bundle": "^4.0",
+        "symfony/http-foundation": "^4.0",
+        "symfony/http-kernel": "^4.0",
+        "symfony/monolog-bundle": "^3.2",
+        "symfony/routing": "^4.0",
+        "symfony/yaml": "^4.0"
+    },
```
giorgiosironi, line 18: I assume this is pure Symfony 4. Wonder about the size of `vendor/` and how it compares to the current Silex version it is substituting?

api-dummy/config/packages/prod/monolog.yaml
```diff
@@ -0,0 +1,16 @@
+monolog:
```
giorgiosironi, line 1: There is an assumption about which are the standard environments supported: `dev`, `prod`, what else?

api-dummy/public/index.php
```diff
@@ -0,0 +1,30 @@
...
+use Symfony\Component\Debug\Debug;
+use Symfony\Component\HttpFoundation\Request;
+
+require __DIR__.'/../vendor/autoload.php';
+
+$env = $_SERVER['APP_ENV'] ?? 'dev';
+$debug = (bool) ($_SERVER['APP_DEBUG'] ?? ('prod' !== $env));
+
+if ($debug) {
+    umask(0000);
```
giorgiosironi: similar to the console's `umask`

api-dummy/symfony.lock
```diff
@@ -0,0 +1,98 @@
+{
```
giorgiosironi, line 1: Symfony Flex is a candidate for knowledge sharing

api-dummy/bin/console
thewilkybarkid: Just Symfony boilerplate.

api-dummy/config/packages/prod/monolog.yaml
thewilkybarkid: Depends on the infrastructure. `ci`/`test`; we could end up with `end2end` putting all the distributed Libero projects together. Maybe a `demo` that we host?

giorgiosironi: The name of environments are good, we should only take care:
- not to put in assumptions about hostnames which we usually don't
- leave open the possibility to add new environments without forking the code (e.g. extending the Docker image and adding files)

api-dummy/.docker/nginx-server.conf
giorgiosironi: so I'd remove the two `error_log` and `access_log` directives

api-dummy/Dockerfile
thewilkybarkid: Sounds out of scope for this as a skeleton.

thewilkybarkid:
Anything left to do?

giorgiosironi (approved):
I think this is ready to be merged, key risks are in other areas now

seanwiseman (approved):

PR #8, opened 2018-03-28
Add Journal

thewilkybarkid:
Includes #7. Displays the articles without worrying too much about presentation (ie patterns)/how it converts XML.


journal/src/Dom/Document.php
```diff
@@ -0,0 +1,81 @@
...
+use DOMDocument;
+use DOMNode;
+use DOMXPath;
+use IteratorAggregate;
+use Libero\Journal\Dom\Exception\InvalidDom;
+use function restore_error_handler;
+use function set_error_handler;
+use function strpos;
+
+final class Document implements IteratorAggregate, TraversableNode
```
thewilkybarkid, line 14: This is a copy of a alternative immutable DOM library I've been working on.

journal/src/Dom/Document.php
```diff
@@ -0,0 +1,81 @@
...
+    private $document;
+    private $namespaces = [];
+    private $xPath;
+
+    public function __construct(string $xml, string $defaultNamespace)
+    {
+        $this->document = new DOMDocument();
+        $this->document->preserveWhiteSpace = false;
+
+        set_error_handler(function (int $errno, string $errstr) : bool {
```
giorgiosironi, line 28: :face_with_head_bandage:

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
+version: '3'
+
+services:
+    api-dummy-fpm:
+        build:
+            context: ./api-dummy
+        volumes:
+            - ./api-dummy:/srv/app
```
giorgiosironi: this is a source-based workflow, which is enough for `dev` and `ci` environments (possibly also `demo`). Shouldn't be a problem to extend it later to allow deployment without the source code.

giorgiosironi: The `README` in the root directory needs some basic instruction thought, explaining what `docker-compose up` will do.

journal/src/View/Converter/Block/ParagraphConverter.php
```diff
@@ -0,0 +1,45 @@
...
+            $attributes .= " lang=\"{$object->getAttribute('lang')->toText()}\"";
+            $context['lang'] = $object->getAttribute('lang')->toText();
+        }
+
+        $text = '';
+        foreach ($object as $child) {
+            $text .= $this->converter->convert($child, $context);
+        }
+
+        return <<<EOT
```
giorgiosironi, line 36: the `ViewConverter` instances do not use patterns; is it a temporary simplification or an active decision?

thewilkybarkid: Patterns weren't part of the walking skeleton, just plain HTML.

journal/src/View/Converter/Block/QuoteConverter.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace Libero\Journal\View\Converter\Block;
```
giorgiosironi, line 3: Maybe if this is intended to be a sample or a default journal we should choose prefixes/namespaces like:
- `default-journal` and `Libero\DefaultJournal`
- `sample-journal` and `Libero\SampleJournal`
- `small-journal`, `vanilla-journal`... (characterizing what this subproject does even further)

docker-compose.yml
giorgiosironi: e.g. run some `composer install` commands and have PHP 7.1

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
...
+        volumes:
+            - ./api-dummy/.docker/nginx-server.conf:/etc/nginx/conf.d/default.conf
+            - ./api-dummy/public:/srv/app/public
+        depends_on:
+            - api-dummy-fpm
+    journal-fpm:
+        build:
+            context: ./journal
+        volumes:
+            - ./journal:/srv/app
```
giorgiosironi, line 28: for both projects only `/srv/app/src` (and other relevant subfolders) should be mounted or `var/cache/` starts showing permission problems when the container user does not match the host folder owner

journal/src/View/Converter/Block/FigureConverter.php
```diff
@@ -0,0 +1,30 @@
...
+    {
+        $this->converter = $converter;
+    }
+
+    /**
+     * @param Element $object
+     */
+    public function convert($object, array $context = []) : string
+    {
+        return '<p>[Figure]</p>';
```
giorgiosironi, line 23: can be out of scope to do more than this, but it may be useful to hide into the HTML a dump of the original $object to explore what data is available when looking at a page

giorgiosironi:
I can pull the branch, start all services successfully and navigate #worksonmymachine
As an exploration of this area I don't think a larger scope is needed, the skull of the skeleton is in place.

thewilkybarkid:
:shipit:?

seanwiseman (approved):
👍

PR #9, opened 2018-04-03
Add working Airflow example

seanwiseman:
Will provide a dummy workflow (DAG) with tasks that will simulate some required usage.

These don't have to interact with any other parts of the skeleton at present.

### DIFF
--- a/workflow/airflow/README.md
+++ b/workflow/airflow/README.md
@@ -1,0 +1,44 @@
+# Airflow Example Usage
+
+Dependencies
+------------
+
+* Python >=3.6
+* [pipenv](https://docs.pipenv.org/)
+
+Installation
+------------
+This will setup your virtual environment and install the required dependencies:
+
+`pipenv --three install`
+
+Set the `AIRFLOW_HOME` environment variable in the `.env` file to your absolute path:
+
+e.g. `AIRFLOW_HOME=~/...../walking-skeleton/workflow/airflow`
+
+Activate your virtual environment:
+
+`pipenv shell`
+
+Initialize the database:
+
+`airflow initdb`
+
+This step will setup an `airflow.db` `sqlite` file, your `airflow.cfg` file and your `unittests.cfg` file.
+ 
+Usage
+-----
+
+Activate your virtual environment:
+
+`pipenv shell`
+
+Start a scheduler instance in it's own process:
+
+`airflow scheduler`
+
+Start a webserver instance in it's own process: (not required but provides a nice UI)
+
+`airflow webserver -p 8080`
+
+Browsing to [http://localhost:8080/](http://localhost:8080/) will now take you to the admin console.
--- a/workflow/airflow/README.md
+++ b/workflow/airflow/README.md
@@ -41,4 +41,4 @@ Start a webserver instance in it's own process: (not required but provides a nic
 
 `airflow webserver -p 8080`
 
-Browsing to [http://localhost:8080/](http://localhost:8080/) will now take you to the admin console.
+Browsing to [http://localhost:8080/](http://localhost:8080/) will now take you to the [admin console](https://airflow.apache.org/ui.html).
--- a/workflow/airflow/README.md
+++ b/workflow/airflow/README.md
@@ -51,7 +51,7 @@ There are multiple ways to trigger a DAG manually.
 
 [CLI](https://airflow.apache.org/cli.html#trigger_dag)
 
-`airflow trigger_dag foo_dag`
+`airflow trigger_dag foo_python_operator`
 
 [UI](https://airflow.apache.org/ui.html#)
 
--- a/workflow/airflow/README.md
+++ b/workflow/airflow/README.md
@@ -41,4 +41,18 @@ Start a webserver instance in it's own process: (not required but provides a nic
 
 `airflow webserver -p 8080`
 
-Browsing to [http://localhost:8080/](http://localhost:8080/) will now take you to the [admin console](https://airflow.apache.org/ui.html).
+Browsing to [http://localhost:8080/](http://localhost:8080/) will now take you to the [admin console](https://airflow.apache.org/ui.html)foo_dag.py.
+
+
+Trigger a DAG
+-----------
+
+There are multiple ways to trigger a DAG manually.
+
+[CLI](https://airflow.apache.org/cli.html#trigger_dag)
+
+`airflow trigger_dag foo_dag`
+
+[UI](https://airflow.apache.org/ui.html#)
+
+On the DAGS view, you can locate a required DAG type and select `Trigger Dag` via the Links column.
--- a/workflow/airflow/README.md
+++ b/workflow/airflow/README.md
@@ -10,7 +10,7 @@ Installation
 ------------
 This will setup your virtual environment and install the required dependencies:
 
-`pipenv --three install`
+`pipenv --python 3.6.3 install`
 
 Set the `AIRFLOW_HOME` environment variable in the `.env` file to your absolute path:
 
--- a/workflow/airflow/dags/local_file_sensor_dag.py
+++ b/workflow/airflow/dags/local_file_sensor_dag.py
@@ -1,0 +1,94 @@
+import os
+
+import airflow
+from airflow import DAG
+from airflow.operators.python_operator import PythonOperator
+from airflow.operators.sensors import BaseSensorOperator
+from datetime import timedelta
+
+
+class FileWatcherOperator(BaseSensorOperator):
+    """
+    Triggers a file system check. Parses a file if present.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Initialize a FileWatcherOperator which should wait for a file to be present in a target directory.
+
+        :param args:
+        :param kwargs:
+        """
+        self.directory_path = kwargs.get('directory_path')
+        super(FileWatcherOperator, self).__init__(*args, **kwargs)
+
+    def poke(self, context, *args, **kwargs):
+        """
+        See if a file is present in a given directory.
+
+        :param context: Context dictionary forwarded from the DAG;
+        expects to contain context.dag and context.task
+        :type context: dict
+        :param args: not expected
+        :param kwargs: not expected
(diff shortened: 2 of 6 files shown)


workflow/airflow/README.md
```diff
@@ -0,0 +1,44 @@
...
+Usage
+-----
+
+Activate your virtual environment:
+
+`pipenv shell`
+
+Start a scheduler instance in it's own process:
+
+`airflow scheduler`
```
giorgiosironi, line 38: are there worker processes to run or does the `scheduler` work by itself?

workflow/airflow/dags/local_file_sensor_dag.py
```diff
@@ -0,0 +1,94 @@
+import os
+
+import airflow
+from airflow import DAG
+from airflow.operators.python_operator import PythonOperator
+from airflow.operators.sensors import BaseSensorOperator
+from datetime import timedelta
+
+
+class FileWatcherOperator(BaseSensorOperator):
```
giorgiosironi, line 10: suppose I want to write a test for this class. Which options do I have? Unit tests by stubbing lots of methods or integration tests using a real airflow process?

workflow/airflow/dags/local_file_sensor_dag.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    Triggers a file system check. Parses a file if present.
```
giorgiosironi: This doesn't seem to actually parse the file, only to get its name right?

workflow/airflow/dags/local_file_sensor_dag.py
```diff
@@ -0,0 +1,94 @@
...
+        :param args: not expected
+        :param kwargs: not expected
+        :return: whether the poke method has succeeded or not, 0 or 1
+        :rtype: int
+        """
+        path = os.path.join(os.getcwd(), self.directory_path)
+        dir_listing = os.listdir(path)
+
+        if len(dir_listing) > 0:
+            context['ti'].xcom_push(key='file_path', value=os.path.join(path, dir_listing[0]))
```
giorgiosironi, line 41: so this is how operators communicate. Is `file_path` available only in the directly downstream operator, or does it become more "global" in further operators down the workflow?

workflow/airflow/dags/local_file_sensor_dag.py
```diff
@@ -0,0 +1,94 @@
...
+
+def parse_file(*args, **kwargs):
+    """Simply parse the first file in a directory and return it's contents.
+
+    :param args:
+    :param kwargs:
+    :return:
+    """
+    # get file path value from passed in context
+    file_path = kwargs['ti'].xcom_pull(task_ids='file_watcher', key='file_path')
```
giorgiosironi, line 58: `task_id[s]` seems to be dinamically defined for `FileWatcherOperator`, while here it is hardcoded. What is the standard way to do it?

workflow/airflow/dags/local_file_sensor_dag.py
```diff
@@ -0,0 +1,94 @@
...
+
+    with open(file_path) as f:
+        return f.read()
```
giorgiosironi, line 61: so once a value is returned, where does it go? In general we tend to have "void" activities that have some side-effect on an external resource (push some files somewhere, call an API, put a message into a queue) so what use cases are return values good for?

workflow/airflow/dags/local_file_sensor_dag.py
```diff
@@ -0,0 +1,94 @@
...
+
+# the second task is going to wait for the watcher to return true before starting
+file_parser = PythonOperator(task_id='file_parser',
+                             provide_context=True,
+                             python_callable=parse_file,
+                             dag=dag)
+
+
+# set the file_parser to come after the watcher
+file_parser.set_upstream(watcher)
```
giorgiosironi, line 94: Too soon to know, but I was thinking about breaking down workflows into smaller pieces due to an external asynchronous dependency (e.g. put something in bot-lax-adaptor's queue, wait for it to come back with a result). In hindsight I'd model those as a synchronous API call to simplify it.

workflow/airflow/README.md
seanwiseman: With the current setup (basically only suitable for local testing and very lightweight use) it uses a `SequentialExecutor` by default which will only run task instances sequentially (no workers needed). In production, we would probably use its `CeleryExecutor` which will require `rabbitmq` and `redis` and then it would need its own workers for parallel task processing.

workflow/airflow/dags/local_file_sensor_dag.py
seanwiseman: This was a mistake. Should have been taken out during a small refactor. Now removed as yes it does not parse the file.

workflow/airflow/dags/local_file_sensor_dag.py
seanwiseman: Yes their XCOM system is how you can communicate between steps in a workflow. It's effectively a key-value store in which a workflow instance will have a namespace.

You can access the `return_value` of a given operation by just referring to a `task_id` or you can push and pull via a `key` which I found more flexible and clearer.

workflow/airflow/dags/local_file_sensor_dag.py
seanwiseman: In the `FileWatcherOperator` I am pushing a value onto the store so you don't need to define `task_ids`, it knows what task you are pushing from.

In the `parse_file` function I am saying get me the value for key `file_path` that was populated by the `file_watcher` operator:

`file_path = kwargs['ti'].xcom_pull(task_ids='file_watcher', key='file_path')`

So when pulling values you are required to define a value for `task_ids` otherwise you get:

`TypeError: xcom_pull() missing 1 required positional argument: 'task_ids'`

If you pass `None` to `task_ids` it will just look for the first key `file_path` populated by any of your tasks in that instance (not recommended as could lead to unpredictable results).

If you pass a list of operator names to `task_ids` it will return a tuple of multiple values, in this example, you might have lots of operators pushing file paths and you may want to parse them all at the end (this is contrived but you get what I mean). e.g.

`file_paths = kwargs['ti'].xcom_pull(task_ids=['watcher_1', 'watcher_2', 'watcher_3'], key='file_path')`
```
>>> file_paths
('/path/one', '/path/two', '/path/three')
```

workflow/airflow/dags/local_file_sensor_dag.py
seanwiseman: In this case its available for other operators to access via XCOM, but we could just as well have executed a side effect within the call and not bothered with returning the value. They can just be "void" activities/tasks if you want them to be.

workflow/airflow/dags/local_file_sensor_dag.py
seanwiseman: There are many different ways to go about workflow (DAG) orchestration with the functionality `airflow` offers. Once we have a set of steps for a workflow we actually want to try and model we can try a few different implementations.

workflow/airflow/README.md
giorgiosironi: a `CeleryExecutor` should be part of the walking skeleton, but not of this PR

workflow/airflow/dags/local_file_sensor_dag.py
giorgiosironi: > a key which I found more flexible and clearer

:+1:

workflow/airflow/dags/local_file_sensor_dag.py
seanwiseman: In the next PR I will add some components for building DAGs (small module of reusable operators) along with their own tests. 

Unit tests won't need that much stubbing, after stubbing a DAG, the config and context passed around they are just Python functions and classes.

I would also have some integration tests for confidence that an entire workflow does what we think. There is a fairly lightly documented `TEST_MODE` you can put an `airflow` instance into which might help, would need to investigate this feature further.

workflow/airflow/dags/local_file_sensor_dag.py
giorgiosironi: > If you pass None to task_ids it will just look for the first key file_path populated by any of your tasks in that instance (not recommended as could lead to unpredictable results).

Unpredictable, true, but if used like the elife-bot's session these values are populated once and never overwritten; so in that use case it could work (but only loosely enforced).

seanwiseman:
I will leave the `docker` setup and configuration additions for the next PR when the production set of components will be used in favour of in memory and `sqlite`.

workflow/airflow/README.md
```diff
@@ -0,0 +1,58 @@
...
+------------
+
+* Python >=3.6
+* [pipenv](https://docs.pipenv.org/)
+
+Installation
+------------
+This will setup your virtual environment and install the required dependencies:
+
+`pipenv --three install`
```
thewilkybarkid: I had to use `pipenv --python 3.6.3 install`, otherwise it found 3.5 for unknown reasons.

seanwiseman: If that was your system default for 3 then that's understandable.

I will update the readme to account for this.

workflow/airflow/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+There are multiple ways to trigger a DAG manually.
+
+[CLI](https://airflow.apache.org/cli.html#trigger_dag)
+
+`airflow trigger_dag foo_dag`
+
+[UI](https://airflow.apache.org/ui.html#)
+
+On the DAGS view, you can locate a required DAG type and select `Trigger Dag` via the Links column.
```
thewilkybarkid: `foo_dag` called `foo_python_operator` here.

seanwiseman: 👍 (file name instead of `id` was used)

thewilkybarkid (approved):
Seems ok for now, something to continue to explore (as part of the general choreography–orchestration question).

seanwiseman:
@giorgiosironi anything else to add?

giorgiosironi (approved):
Was making scattered comments, but this is good to go

PR #10, opened 2018-04-12
Add more languages/scripts

thewilkybarkid:
To help us think more about languages/scripts I've added (with the help of Google Translate) some more content 

- Article in language not available in the UI (Portuguese; so currently not visible at all)
- Arabic (right to left) UI/article
- Japanese UI (quite different script to Latin)


journal/src/View/Converter/Block/AsideConverter.php
```diff
@@ -27,18 +30,24 @@ public function convert($object, array $context = []) : string
         if ($id = $object->getAttribute('id')) {
             $attributes .= " id=\"{$id->toText()}\"";
         }
-        if ($object->getAttribute('lang') ?? $context['lang'] !== $context['lang']) {
-            $attributes .= " lang=\"{$object->getAttribute('lang')->toText()}\"";
+        if ($object->getAttribute('lang') && $object->getAttribute('lang')->toText() !== $context['lang']) {
```
thewilkybarkid, line 33: This setup is repeated a lot, but I'm ignoring it for now.

journal/src/View/Converter/Block/AsideConverter.php
```diff
@@ -27,18 +30,24 @@ public function convert($object, array $context = []) : string
...
             $context['lang'] = $object->getAttribute('lang')->toText();
+            $dir = $this->translator->trans('direction', [], null, $context['lang']);
+
+            $attributes .= " lang=\"{$context['lang']}\"";
+            $attributes .= " dir=\"{$dir}\"";
```
thewilkybarkid, line 38: Ideally wouldn't repeat if the context is already the same direction, but I'm ignoring it for now.

journal/translations/messages.ar.yaml
```diff
@@ -0,0 +1,16 @@
...
+            es: 'هذه المقالة مكتوبة باللغة الإسبانية.'
+            ja: 'هذه المقالة مكتوبة باللغة اليابانية.'
+            ru: 'هذه المقالة مكتوبة باللغة الروسية.'
+    articles: 'مقالات'
+    digest: 'موجز'
+    home: 'رئيسية'
+    language_choice: 'اختار اللغة:'
+    no_articles: 'لا توجد مقالات متاحة.'
+    title: 'عرض Libero'
+direction: rtl
```
thewilkybarkid, line 16: Doesn't seem like a great way to set the direction (what happens if some content doesn't have a translation file?), but ignoring it for now.

api-dummy/public/schemas/libero/core/inline.rng
```diff
@@ -27,6 +27,13 @@
         </element>
     </define>
 
+    <define name="libero.text.foreign">
+        <element name="foreign">
```
giorgiosironi, line 31: not sure `foreign` is the right term here, it could be a second or bilingual language for the reader? Works as a placeholder for now

giorgiosironi (approved):

api-dummy/public/schemas/libero/core/inline.rng
thewilkybarkid: Foreign as in different to the surrounding text ([borrowed from TEI](http://www.tei-c.org/release/doc/tei-p5-doc/en/html/ref-foreign.html)). But yeah, these are all placeholders anyhow.

seanwiseman (approved):
🌎 👍

PR #11, opened 2018-04-16
Libero flow: A POC Workflow Management System

seanwiseman:
A minimal custom solution performing the same functionality as other off the shelve solutions but with the bare bones required. At the very least, this is being done as an insightful comparison for exploratory purposes.

Refer to the README (WIP) for an implementation overview.


workflow/libero_flow_poc/docker-compose.yml
```diff
@@ -0,0 +1,88 @@
...
+      - rabbitmq
+
+  redis:
+    image: 'redis:3.0-alpine'
+    ports:
+      - '6379:6379'
+    volumes:
+      - 'redis:/data'
+
+  rabbitmq:
```
giorgiosironi, line 77: Just pointing out that this RabbitMQ seems to be internal to the project, rather than being used to communicate with other components. Therefore it doesn't need to conform to the ADRs about the message bus (for example the message schema), though the fact that it is RabbitMQ makes it easier to operate since its setup will be similar to what is used in the message bus itself.

workflow/libero_flow_poc/libero_flow/decider.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def send_decision_message(decision: Dict):
+    """create and send decision task message.
+
+    :param decision: dict
+    :return:
+    """
+    with get_channel() as channel:
+        message = {
```
giorgiosironi: same as previous comment - no need to force messages into the bus schema if these messages are internal to this component.

It may be useful for the bus to separately receive events about what happens inside the worfklow of course.

workflow/libero_flow_poc/docker-compose.yml
seanwiseman: The possible benefit of using the global bus would be other things could react to the events, but yes by design at present this has its own.

workflow/libero_flow_poc/libero_flow/decider.py
seanwiseman: 👍

workflow/libero_flow_poc/libero_flow/flow/workflow/tests/test_api/test_activity_api.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from workflow.models import Activity
+
+
+@pytest.mark.django_db
+def test_can_create_activity_via_post(admin_client, valid_activity_data):
+    response = admin_client.post('/workflows/api/v1/activities/', data=valid_activity_data)
```
giorgiosironi, line 8: what is the use case for creating an activity via API outside of a larger workflow? Would it be able to run on its own?

workflow/libero_flow_poc/libero_flow/scheduler.py
```diff
@@ -0,0 +1,231 @@
...
+    create_workflow(data['name'], data['input_data'])
+
+
+@setup_exchanges_and_queues
+def main():
+    with get_channel() as channel:
+        print('Scheduler running...')
+        print(' [*] Waiting for Messages. To exit press CTRL+C')
+
+        channel.basic_consume(message_handler(activity_result_handler), queue=ACTIVITY_RESULT_QUEUE, no_ack=False)
```
giorgiosironi: are these 3 calls concurrent? Or running one after the other?

giorgiosironi: Would seem strange to e.g. waiting on activity results while there are other decision results to show. This is probably just a debugging method however.

giorgiosironi (commented):
Waiting on a README for more architectural details, so far it would be interesting to understand which part of the code is needed in any implementation and which part is substituting Airflow.

workflow/libero_flow_poc/docker-compose.yml
```diff
@@ -0,0 +1,88 @@
...
+            "
+    depends_on:
+      - rabbitmq
+
+  redis:
+    image: 'redis:3.0-alpine'
+    ports:
+      - '6379:6379'
+    volumes:
+      - 'redis:/data'
```
giorgiosironi, line 75: I didn't understand what Redis was used for?

workflow/libero_flow_poc/docker-compose.yml
giorgiosironi: The more this is hidden, the easier it is to change without other projects possibly breaking

workflow/libero_flow_poc/libero_flow/scheduler.py
seanwiseman: They are executed on one thread (for this POC) but are just callback assignments. Meaning you can receive messages from all 3 queues independently, but I can only execute one callback at a time.

As the callback functions don't hold state it should not matter, you could implement multi threading/processing or start x2 schedulers and you could then execute two callbacks at a time etc.

workflow/libero_flow_poc/docker-compose.yml
seanwiseman: The session layer to store workflow instance values, though I haven't committed the `session_store` code yet 👾 as still tweaking.

workflow/libero_flow_poc/libero_flow/flow/workflow/tests/test_api/test_activity_api.py
seanwiseman: The `scheduler` uses that to create each initial activity state and the `Activity` status is
updated using a `PATCH` on the same endpoint but providing its id.

This means I did not need to give the other processes database access, could be run anywhere and just use the API.

workflow/libero_flow_poc/docker-compose.yml
seanwiseman: Yes good point.

seanwiseman:
This does not need `AirFlow` at all, it is just intended to be a stripped back alternative building only the required features.

workflow/libero_flow_poc/libero_flow/flow/workflow/tests/test_api/test_activity_api.py
giorgiosironi: I'm all for exposing APIs to the other services rather than giving access to the internals like a database (or a bucket or whatever resource is being used), although in the case of these containers they all come from the same source code so I would expect them to run in the same deployment, maybe not on the same node or Kubernetes pod but definitely inside the same VPC. So giving access to the database should not be a big deal for these processes, while it may be useful to go through the API for extensions (we don't know yet what those would do or how they would be implemented).

seanwiseman: Agreed, if need be, with the current project structure you could easily give them an interface to the database. Though as you have mentioned, currently I can do other things when the API requests are processed which is handy which you may not know about if the tables were just updated directly.

giorgiosironi:
Code is a useful feedback loop here, forcing dependencies such as databases and libraries to come out.
I think a comparison needs also to be made with people around a whiteboard, at an higher level of abstraction. Two diagrams showing the moving parts of both the solutions?

seanwiseman:
Definitely, as with `AirfFlow` and `SWF`, its easy to get lost in the different parts required and how they interact.

giorgiosironi:
The `workflow/libero_flow` seems to be a leftover as everything is in the other folders, can it be removed?

giorgiosironi:
The `ui` image has some problem being built on my machine, `npm install` failing some DNS resolution:
```
npm ERR! code EAI_AGAIN
npm ERR! errno EAI_AGAIN
npm ERR! request to https://registry.npmjs.org/array-slice/-/array-slice-0.2.3.tgz failed, reason: getaddrinfo EAI_AGAIN registry.npmjs.org
:443
```
Anyone else encountered this problem?

seanwiseman:
I can't seem to replicate on my mac. 

Does it successfully install anything using `npm`? if not can you run a container using the `node:8.7-alpine` base and just try and `npm install` something?

giorgiosironi:
Got past that, was probably due to the network.

I have opened http://localhost:4200/ which redirects to http://localhost:4200/workflows but an error is reported in the `postgres` container:
```
 postgres_1   | STATEMENT:  SELECT "workflow_workflow"."instance_id", "workflow_workflow"."name", "workflow_workflow"."status", "workflow_workflow"."created", "workflow_workflow"."start_timestamp", "workflow_workflow"."end_timestamp", "workflow_workflow"."input_data", "workflow_workflow"."config" FROM "workflow_workflow" ORDER BY "workflow_workflow"."created" DESC LIMIT 21
```

workflow/libero_flow_poc/README.md
```diff
@@ -0,0 +1,252 @@
...
+
+Start a Workflow by http request:
+```python
+
+import requests
+
+api_root = 'http://localhost:8000'  # this will depend on your workflow API server deployment 
+
+payload = {"name": "DummyWorkflow", "input_data": {}}
+url = api_root + '/workflows/api/v1/startworkflow/'
```
giorgiosironi, line 168: this seems to be `start-workflow`:
```
curl -v http://localhost:8000/workflows/api/v1/start-workflow/ -d 'name=DummyWorkflow&input_data[key]=value'
{"message":"Start Workflow request sent"}
```
However after doing that, nothing is happening for me? No activity in http://localhost:4200/workflows or in Docker logs.

seanwiseman: I suspect you do not have any workflows defined. `DummyWorkflow` does not exist by default, it was just an example in the README.

You can create the `DummyWorkflow` as shown in the README, or just create a new one.

Then just add any `activities` you define in the workflow.

workflow/libero_flow_poc/libero_flow/state_utils.py
```diff
@@ -0,0 +1,86 @@
...
+import requests
+
+from libero_flow.conf import (
+    ACTIVITY_API_URL,
+    EVENT_API_URL,
+    WORKFLOW_API_URL,
+)
+
+
+WORKFLOW_DECISION_SCHEDULED = 'WorkflowDecisionScheduled'
```
giorgiosironi, line 12: it seems `WorkflowDecisionStarted` and `WorkflowDecisionFinished` are missing, for completeness. Not necessarily to be exposed to the event bus, but part of the diagnostics here? (`DEBUG`-level logging for example)

workflow/libero_flow_poc/libero_flow/worker.py
```diff
@@ -0,0 +1,157 @@
...
+
+        update_activity_status(activity_id, status='In Progress')
+
+        loader = FlowLoader()
+        activity_state = get_activity_state(activity_id=activity_id)
+        activity_class = loader.get_activity(activity_state['name'])
+
+        if activity_class:
+            send_workflow_event(workflow_id=activity_state['workflow'],
+                                event_type=WORKFLOW_ACTIVITY_STARTED)
```
giorgiosironi, line 60: these events should definitely contain `activity_class` or some normalization of it

workflow/libero_flow_poc/libero_flow/worker.py
```diff
@@ -0,0 +1,157 @@
...
+                                      config=activity_state['config'],
+                                      session=session)
+            result['result'] = activity.do_activity()
+
+            send_workflow_event(workflow_id=activity_state['workflow'],
+                                event_type=WORKFLOW_ACTIVITY_FINISHED)
+        else:
+            result['result'] = 'no-activity-found'
+
+    except Exception as err:
```
giorgiosironi, line 74: it would be useful to turn exceptions into a `WorkflowActivityFailed` event (or similar)

workflow/libero_flow_poc/libero_flow/scheduler.py
```diff
@@ -0,0 +1,217 @@
...
+    """
+    with get_channel() as channel:
+        message = get_base_message()
+        message['aggregate']['service'] = 'flow-scheduler'
+        message['aggregate']['name'] = 'schedule-workflow-activity'
+        message['type'] = 'schedule-activity'
+        message['data'] = {'activity_id': activity_id}
+
+        channel.basic_publish(exchange=SCHEDULED_ACTIVITY_EXCHANGE,
+                              routing_key=SCHEDULED_ACTIVITY_QUEUE,
```
giorgiosironi: Am I getting RabbitMQ exchanges wrong or are all the `routing_key`s not needed? It should suffice to publish to the exchange which will fan out to its (only?) queue.

workflow/libero_flow_poc/libero_flow/decider.py
```diff
@@ -0,0 +1,196 @@
...
+
+    elif workflow['status'] == PENDING:
+        decision['decision'] = 'start-workflow'
+
+    elif workflow['status'] == IN_PROGRESS:
+        # do some real work
+        activities_to_schedule = []
+        decision['decision'] = 'schedule-activities'
+
+        for activity in workflow['activities']:
```
giorgiosironi: why are we going through multiple activities here? I thought only one would be started at a time... unless this is to keep as much parallelism as possible (like branching 3 activities to be run simultaneously)

seanwiseman: Yes, if a workflow has multiple independent activities then they will be started at the same time as they do not depend on other activity chaining or values.

workflow/libero_flow_poc/libero_flow/state_utils.py
seanwiseman: I was on the fence when implementing the events whether to add these. As the decider has a consistent algorithm unaffected by user-defined functions (like activities are), unless the actual decider fails, the output and execution time pretty static and predictable.

But as you say, these can just be filtered anyway so I will add them.

workflow/libero_flow_poc/libero_flow/worker.py
seanwiseman: I agree. It should probably be a generic field, something like 'info' as it's not just activity based events being logged, but also workflow and decision events.

It would require touching a few parts (the db model, api, the func calls) so is probably only worth touching if we are to expand on this experiment past this PR.

workflow/libero_flow_poc/libero_flow/scheduler.py
seanwiseman: Yes from my understanding, you should be able to publish to the exchange.

Though the `routing_key` is a required field:

`basic_publish(exchange, routing_key, body, properties=None, mandatory=False, immediate=False)`

I will investigate further for future use cases, but in this instance, removing the `routing_key` raises:

`TypeError: basic_publish() missing 1 required positional argument: 'routing_key'`

And passing `None` gives you:
`'A non-string value was supplied for self.routing_key'`

giorgiosironi: The `fanout` exchange should ignore routing keys, so there should be some conventional value like `*` to pass in

seanwiseman: 👍  I will investigate further.

seanwiseman: Sorted. The old empty string comes to the rescue.

giorgiosironi:
Haven't got much time today but I'll try the Airflow version tomorrow, should I use this or https://github.com/libero/walking-skeleton/pull/15 ?

seanwiseman:
Yes use [#15](https://github.com/libero/walking-skeleton/pull/15) for the Airflow testing. I am just sorting the `docker-compose` file for it then it will be ready to go.

seanwiseman:
Anything else to add to this @giorgiosironi ?

giorgiosironi:
Nothing to add to the previous comments, up to you to shelve or merge this given it's not the first option for implementation.

seanwiseman:
Will close the PR and let it exist on this branch encase its needed for any kind of reference.

PR #13, opened 2018-04-17
Select two scenarios for walking skeleton

giorgiosironi:
Imagining an article in the correct format (the expected XML schema), without any figures or multilingual concerns.

The two scenarios should force the creation of the following components:

- article store
- bot
- dashboard
- journal

The other scenarios can then be prioritized and implemented after the skeleton.


thewilkybarkid:
Do we need previews here?

giorgiosironi:
You can think of preview as a separate feature, so I separated into another scenario, that can be prioritized independently. I'm keen on keeping the ingest and publish separation because it forces the dashboard to talk back to the bot in some way.

thewilkybarkid:
Hmm, the Journal part of preview is small, I was more talking about the concept.

If preview/unpublished is included, probably don't need to actually hook up a button in the dashboard UI? If it's a HTTP API, cURL should suffice?

giorgiosironi:
The meaning was that if the dashboard is not read only, but it's enacting actions upon the bot; therefore that arrow between the two boxes it's interesting to explore. We can trade-off a bit more coverage and assume we are able to configure a request down the line, just exposing a HTTP API from the bot and calling it from manually or from the first end2end test.

giorgiosironi:
Getting closer to the first two selected scenario, should this be merged now?

publication.feature
```diff
@@ -0,0 +1,46 @@
...
+  @future
+  Scenario: Figures versions
+    Given an article X with 2 figures
+    When I ingest and publish article X
+    Then I should see version 2 of article X on the dashboard in published state
+    And I should see version 2 of article X on journal
+
+  @future
+  Scenario: Input formats
+    Given an article X in JATS format
```
thewilkybarkid: This is out of scope.

thewilkybarkid:
Going to tweak the language a bit.

publication.feature
```diff
@@ -0,0 +1,46 @@
...
+
+  @future
+  Scenario: Multilingual articles
+    Given an ingested article X in languages en and pr
+    When I ingest and publish article X
+    Then I should see article X on journal
+    And I should see the article X in pr on journal
+
+  @future
+  Scenario: Figures versions
```
thewilkybarkid: Seems redundant?

giorgiosironi: The title is wrong, it should just be `Figures`

giorgiosironi:
Amended the scenarios, feel free to change the wording.

publication.feature
```diff
@@ -0,0 +1,48 @@
+Feature: Publication of an article
+
+  Scenario: Make an article version ready to publish
+    Given version 1 of article 1234 is in English
```
thewilkybarkid, line 4: 'version _x_ of' everywhere is getting a bit wordy...

thewilkybarkid: As is the language

giorgiosironi: Could use a mnemonic code like `article 1234v1`

publication.feature
```diff
@@ -1,40 +1,48 @@
...
+    And I can read version 2 of article 1234 on Journal
 
   @future
-  Scenario: Figures
-    Given an article X with 2 figures
-    When I ingest and publish article X
-    Then I should see article X on the dashboard in published state
-    And I should see article X on journal
+  Scenario: Publish an article version in multiple languages
+    Given version 1 of article 1234 is in English and Portuguese
```
giorgiosironi: It would be pretty standard in my opinion not to specify a language (but having a default of `English`) in all scenarios but this, which deals with multiple languages; those details do not change the behavior of the existing scenarios

giorgiosironi:
Sneakily reassigned after in-person discussion, for refinement.

thewilkybarkid (approved):
I'm trying not to worry any more about the language (which is ironic for Gherkin), since it's just the skeleton.

PR #15, opened 2018-04-23
Add ingest article xml DAG to Airflow Demo

seanwiseman:
Adding a slightly more realistic dag to show how we can:

- pass data to a dag on initialization
- store some values between tasks
- emit events at key points of the dag execution to allow other services to know dag state


workflow/airflow_libero/README.md
```diff
@@ -0,0 +1,116 @@
...
+* Python >=3.6
+* [pipenv](https://docs.pipenv.org/)
+
+Installation
+------------
+This will setup your virtual environment and install the required dependencies:
+
+`pipenv --python 3.6.3 install`
+
+Set the `AIRFLOW_HOME` environment variable in the `.env` file to your absolute path:
```
giorgiosironi: is this `workflow/airflow/.env`? Because it's in another directory and it's not used by `docker-compose.yml` :thinking:

workflow/airflow_libero/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+`docker-compose up`
+
+Start the event listener:
+
+`pipenv run python run_event_bus_listener.py`
+
+Start a scheduler instance in it's own process:
+
+`pipenv run airflow scheduler`
```
giorgiosironi, line 44: seems to be missing some tables for me
```
...
  File "/home/giorgio/.local/share/virtualenvs/airflow_libero-4mEf_ytu/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 507, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: job [SQL: 'INSERT INTO job (dag_id, state, job_type, start_date, end_date, latest_heartbeat, executor_class, hostname, unixname) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)'] [parameters: (None, 'running', 'SchedulerJob', '2018-04-25 09:18:29.710916', None, '2018-04-25 09:18:29.710920', 'SequentialExecutor', 'Newton', 'giorgio')]
```

workflow/airflow_libero/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+Start a webserver instance in it's own process: (not required but provides a nice UI)
+
+`pipenv run airflow webserver -p 8080`
```
giorgiosironi, line 48: same for this process
```
...
  File "/home/giorgio/.local/share/virtualenvs/airflow_libero-4mEf_ytu/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 507, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: connection [SQL: 'SELECT connection.conn_id AS connection_conn_id \nFROM connection GROUP BY connection.conn_id'] (Background on this error at: http://sqlalche.me/e/e3q8)
```

workflow/airflow_libero/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+Activate your virtual environment:
+
+`pipenv shell`
+
+Initialize the database:
+
+`airflow initdb`
+
+This step will setup an `airflow.db` `sqlite` file, your `airflow.cfg` file and your `unittests.cfg` file.
```
giorgiosironi, line 29: files are correctly created for me

seanwiseman:
(was meant to have posted this earlier, never clicked `Comment`)

You can now run an example ingest article xml dag @giorgiosironi .

This displays the following functionality:
- Using a custom `EventEmittingPythonOperator` for tasks, this allows events destined for an event bus to be emitted at key points of execution.
- Triggering dags via cli or http.
- Receiving events for dag runs.

Currently I have only included `rabbitmq` in the `docker-compose.yml` file, the rest just needs to be spun up locally for this example test. I have got about 80% of the way through getting a full deployment working via `docker-compose` but it needs some work and was not the sole purpose of this PR.

giorgiosironi:
I'm not opposed to use off-the-shelf solution in general; so far, Airflow looks like a distributed cron for large batch jobs (e.g. repeated data imports) more than for instantiable, independent similar jobs. There is the added complexity of its various concepts, for example the admin UI is vast with only a few options being usable. Will now look inside the code.

workflow/airflow_libero/dags/ingest_article_xml_dag.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def _download_assets(*args, **kwargs):
+    """Download some static assets.
+
+    :return: str
+    """
+
+    input_data = kwargs['dag_run'].conf.get('input_data')
+    uris = kwargs['ti'].xcom_pull(task_ids=None, key='uris')
```
giorgiosironi, line 77: the `xcom` push and pull doesn't look too bad apart from the strange naming

workflow/airflow_libero/dags/ingest_article_xml_dag.py
```diff
@@ -0,0 +1,125 @@
...
+                                                 python_callable=_extract_asset_uris,
+                                                 dag=dag)
+
+download_assets = EventEmittingPythonOperator(task_id='download_assets',
+                                              provide_context=True,
+                                              python_callable=_download_assets,
+                                              dag=dag)
+
+
+extract_asset_uris.set_upstream(download_article_xml)
```
giorgiosironi, line 124: the complexity of arbitrary DAGs shows here (wrt linear pipelines)

workflow/airflow_libero/dags/ingest_article_xml_dag.py
```diff
@@ -0,0 +1,125 @@
...
+            file_name = uri.split('/')[-1]
+
+            with open(os.path.join(data_dir, file_name), 'wb') as asset_file:
+                asset_file.write(response.content)
+
+
+default_args = {
+    'owner': 'airflow',
+    'depends_on_past': False,
+    'start_date': airflow.utils.dates.days_ago(2),
```
giorgiosironi, line 96: hackish way to get it to run immediately I guess - shows the bias towards scheduling

workflow/airflow_libero/operators.py
```diff
@@ -0,0 +1,40 @@
...
+from airflow.operators.python_operator import PythonOperator
+
+from .event_utils import (
+    get_channel,
+    get_base_message,
+    TASK_EVENT_EXCHANGE,
+)
+
+
+class EventEmittingPythonOperator(PythonOperator):
```
giorgiosironi, line 13: this class is an external integration with RabbitMQ and looks identical to the pure Python case

workflow/airflow_libero/dags/ingest_article_xml_dag.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+download_article_xml = EventEmittingPythonOperator(task_id='download_article_xml',
+                                                   provide_context=True,
+                                                   python_callable=_download_article_xml,
+                                                   dag=dag)
+
+extract_asset_uris = EventEmittingPythonOperator(task_id='extract_asset_uris',
+                                                 provide_context=True,
+                                                 python_callable=_extract_asset_uris,
```
giorgiosironi, line 115: Passing in an object's method rather than a function would be identical right?
Thinking about objects that take collaborators in their constructors, and can then be used as activities.

seanwiseman: Yes as long as the module you are initiating these operators is the one pulling in and setting up the collab classes.

I don't particularly like how you have to define the dags, but they are just classes at the end of the day if you had a lot of custom dags and tasks/operators then a more abstract library structure would be easier to manage and share.

workflow/airflow_libero/Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
```
giorgiosironi, line 1: not sure where they come from, but `numpy` and `pandas` are inside the set of dependencies, adding >100 MB to the virtualenv size

giorgiosironi: They are dependencies it seems?
https://github.com/apache/incubator-airflow/blob/master/setup.py#L247

seanwiseman: It's an absolute beast when it comes to dependencies. Once you run a full setup (celery, postgres, redis + other libs) it gets pretty big. But it does try and integrate with a lot of things.

workflow/airflow_libero/operators.py
seanwiseman: It was the cleanest/quickest way for me to get some events out of the tasks without having to mess with its way of executing. It also has a benefit over a simple decorator as you have access to the dag its being run in.

Would also need to do something similar with the base `DAG` class if we wanted dag events as well (started -- failed -- finished).

seanwiseman:
Anything else to add to this @giorgiosironi ?

giorgiosironi (approved):
Should be merged to get closer to running it with docker-compose

PR #17, opened 2018-05-01
Containerise Airflow

seanwiseman:
The goal is to have a fully containerised `Airflow` installation, one using a `celery-executor` and  another using a `local-executor`.

This was my base I started with: https://github.com/puckel/docker-airflow

I have stripped out a lot of stuff we don't need but there is still room for more and possible refactoring.


seanwiseman:
~~The [`flower`](https://github.com/mher/flower) service is currently failing to start. It's not required as it's just a UI for monitoring the celery cluster(s) but is nice to have.~~

Update: now working after explicitly adding `flower` to depends, even though `Airflow` is meant to add it when defining `celery` as an extra 🤔. Not going to worry to much at present.

seanwiseman:
Ready for some input I think. Currently, the custom DAG for ingesting article XML is not going to work as the imports are off once you put them in containers (this method was only meant to be temporary anyway).

After this is merged I will move the custom operator(s)/DAGs to a separate package and install it as a dependency, allowing the deployment and any custom code to have some breathing space.

workflow/airflow_libero/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+        curl \
+        rsync \
+        netcat \
+        locales \
+    && sed -i 's/^# en_US.UTF-8 UTF-8$/en_US.UTF-8 UTF-8/g' /etc/locale.gen \
+    && locale-gen \
+    && update-locale LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 \
+    && useradd -ms /bin/bash -d ${AIRFLOW_HOME} airflow \
+    && pip install -U pipenv \
+    && pipenv install --system --deploy
```
thewilkybarkid: Should be possible to clean up the build steps, but not worth optimising here.

workflow/airflow_libero/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+
+RUN chown -R airflow: ${AIRFLOW_HOME}
+
+EXPOSE 8080 5555 8793
```
thewilkybarkid: `8080` = UI/REST API?
`5555` = Flower
`8793` = ???

seanwiseman: `worker_log_server_port`
```
# When you start an airflow worker, airflow starts a tiny web server
# subprocess to serve the workers local log files to the airflow main
# web server, who then builds pages and sends them to users. This defines
# the port on which the logs are served. It needs to be unused, and open
# visible from the main web server to connect into the workers.
```

workflow/airflow_libero/Dockerfile
seanwiseman: Yes agreed. I have already stripped out a lot but it could be further refined moving forward.

workflow/airflow_libero/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM python:3.6-slim
```
thewilkybarkid: `python:3.6-jessie` could reduce the amount of build steps? (To make quicker. `slim`/`alpine` would be great for production usage, but that's not needed here.)

seanwiseman: Ok will try it now.

thewilkybarkid:
I'm seeing:

```
scheduler_1  | [2018-05-02 07:59:34,502] {{models.py:288}} ERROR - Failed to import: /usr/local/airflow/dags/ingest_article_xml_dag.py
scheduler_1  | Traceback (most recent call last):
scheduler_1  |   File "/usr/local/lib/python3.6/site-packages/airflow/models.py", line 285, in process_file
scheduler_1  |     m = imp.load_source(mod_name, filepath)
scheduler_1  |   File "/usr/local/lib/python3.6/imp.py", line 172, in load_source
scheduler_1  |     module = _load(spec)
scheduler_1  |   File "<frozen importlib._bootstrap>", line 684, in _load
scheduler_1  |   File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
scheduler_1  |   File "<frozen importlib._bootstrap_external>", line 678, in exec_module
scheduler_1  |   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
scheduler_1  |   File "/usr/local/airflow/dags/ingest_article_xml_dag.py", line 15, in <module>
scheduler_1  |     from airflow_libero.operators import EventEmittingPythonOperator
scheduler_1  | ModuleNotFoundError: No module named 'airflow_libero'
```

seanwiseman:
Re: https://github.com/libero/walking-skeleton/pull/17#issuecomment-385691825

The ingest article xml dag and all its stuff it imports needs moving out into its own package now. Waiting unitil this is merged to move it out.

For the time being, you can just comment out the code in `dags/ingest_article_xml_dag.py`. I left a `foo_python_operator` dag in that you can trigger to make sure the deployment is working.

workflow/airflow_libero/Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+ENV LANG=C.UTF-8
+
+ARG AIRFLOW_VERSION=1.9.0
+ARG AIRFLOW_HOME=/usr/local/airflow
+
+WORKDIR ${AIRFLOW_HOME}
+
+COPY Pipfile ${AIRFLOW_HOME}
+COPY Pipfile.lock ${AIRFLOW_HOME}
+COPY script/entrypoint.sh /entrypoint.sh
```
thewilkybarkid: Change to this file has just triggered a big rebuild. Might be worth moving this and the config to the end, and everything before `pipenv install` above `COPY Pipfile` (a bit of acceptable optimisation!).

thewilkybarkid:
The README needs more updates.

thewilkybarkid:
Managed to run `foo_python_operator` after unpausing (took a while to realise).

`run_event_bus_listener.py` doesn't look to be running yet (so the `rabbitmq` container is redundant)?

workflow/airflow_libero/Dockerfile
seanwiseman: I have moved the `entrypoint.sh` and `config` `COPY`s now 👍 , but moving and splitting the `pipenv` COPY and install commands caused issues with the `WORKDIR` and it could not find the `Pipfile` 🤔 . Leaving that part for now.

seanwiseman:
> run_event_bus_listener.py doesn't look to be running yet (so the rabbitmq container is redundant)?

Well, the `ingest_article_xml` dag was the only thing emitting events, and as that is not going to be functional until moved into its own package, I did not bother running a container for `run_event_bus_listener.py` yet as its going to be moved anyway.

workflow/airflow_libero/Dockerfile
thewilkybarkid: Works for me so will commit.

workflow/airflow_libero/script/entrypoint.sh
```diff
@@ -0,0 +1,96 @@
...
+  AIRFLOW__CORE__LOAD_EXAMPLES=False
+fi
+
+if [ -n "$REDIS_PASSWORD" ]; then
+    REDIS_PREFIX=:${REDIS_PASSWORD}@
+else
+    REDIS_PREFIX=
+fi
+
+wait_for_port() {
```
thewilkybarkid, line 40: Might be a useful to use this pattern elsewhere (eg https://github.com/elifesciences/profiles/blob/ff3ca66bf560bbabc430e2eb09a3172ea17affb3/docker-compose.yml#L44).

thewilkybarkid (commented):
Seems to work. Doesn't cover the whole stack (eg event bus) yet. README could do with updates (eg where to execute `airflow` commands), and `docker-compose.yml` probably needs removing (or move the common stuff there?).

seanwiseman:
[4a311ef](https://github.com/libero/walking-skeleton/pull/17/commits/4a311efed5b547853e736a4b715cedf8d8211efd) 👍 @thewilkybarkid 

Was just looking at what else `jessie` was taking care of, thank you.

seanwiseman:
I have created a base plugin to allow the `ingest_article_xml` dag to work. The custom operator it requires is now imported and included in Airflows ecosystem due to their plugin system. You can execute it from the admin UI.

It's not currently configured to connect to `rabbitmq` (to actually emit the events) as the config part needs extracting out now.

seanwiseman:
Don't think there is anything further to add unless there are suggestions. 

The next step after merging would be moving the event listener bits (it was only POC that events were being sent) and seperating the plugin code.

thewilkybarkid (approved):
Good enough for now.

workflow/airflow_libero/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+WORKDIR ${AIRFLOW_HOME}
+
+RUN set -ex \
+    && apt-get update -yqq \
+    && apt-get install -yqq --no-install-recommends \
+        netcat-openbsd \
+    && useradd -s /bin/bash -d ${AIRFLOW_HOME} airflow \
+    && pip install -U pipenv
+
+COPY Pipfile .
```
giorgiosironi, line 15: can do it in a single `COPY`

workflow/airflow_libero/docker-compose-celery-executor.yml
```diff
@@ -0,0 +1,85 @@
...
+
+    webserver:
+        build:
+          context: .
+        depends_on:
+            - postgres
+            - redis
+        environment:
+            - LOAD_EX=n
+            - FERNET_KEY=46BKJoQYlPPOexq0OhDZnIlNepKFf87WFwLbfzqDDho=
```
giorgiosironi, line 32: I know sometimes large virtualenvs require [a drink](https://en.wikipedia.org/wiki/Fernet)

workflow/airflow_libero/docker-compose-celery-executor.yml
```diff
@@ -0,0 +1,85 @@
...
+            - EXECUTOR=Celery
+        volumes:
+            - './dags:/usr/local/airflow/dags'
+            - './plugins:/usr/local/airflow/plugins'
+        ports:
+            - "8080:8080"
+        command: webserver
+        healthcheck:
```
giorgiosironi, line 40: nice healthcheck, but it can be moved to the Dockerfile with `HEALTHCHECK` so that it is reusable in any docker-compose configuration?

seanwiseman: @giorgiosironi  and I since tried to implement from within the `Dockerfile` but hit an issue easily passing in the `PID_FILE` environment variable from the compose file. 

Left as is for now.

seanwiseman:
@giorgiosironi anything else to add at this point?

workflow/airflow_libero/script/entrypoint.sh
giorgiosironi: It would be very useful to have this function available everywhere, but short of including it in all base images the alternative solution is to render it from a template (https://elifesciences.atlassian.net/browse/ELPP-3600 or similar)

giorgiosironi (approved):

PR #18, opened 2018-05-03
Create dashboard

thewilkybarkid:


dashboard/config/packages/test/web_profiler.yaml
```diff
@@ -0,0 +1,6 @@
+web_profiler:
```
giorgiosironi, line 1: I guess this is default generated configuration, but a profiler and logging into a proof of concept seem to go unused?

giorgiosironi: (while the profiler can be useful in `journal` as performance-sensitive and using parallel requests to the API to build the page)

docker-compose.yml
```diff
@@ -21,6 +21,46 @@ services:
...
+        volumes:
+            - ./dashboard:/srv/app
+            - /srv/app/public/bundles
+            - /srv/app/var
+            - /srv/app/vendor
+        environment:
+            - APP_ENV=dev
+            - APP_SECRET=f6a87c2e031ecadc0d7eb442729435c7
+            - RABBITMQ_URL=amqp://guest:guest@event-bus:5672
+        command: bash -c '/utils/wait-for-port event-bus 5672 && bin/console rabbitmq:consumer events'
```
giorgiosironi: seems easy enough to consume the events. Where to put them to make them available to the UI is a design decision

thewilkybarkid:
Probably enough. Works with an event like:

```json
{
    "eventId": "448278c4-22d6-11e8-b467-0ed5f89f718d",
    "happenedAt": "2018-03-08T11:00:00+00:00",
    "aggregate": {
        "service": "foo",
        "name": "article",
        "identifier": "6551f09a-bd23-4f12-a039-f8fa78ef776b"
    },
    "type": "article.published"
}
```

dashboard/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+USER www-data
+WORKDIR /srv/app
+
+COPY --chown=www-data:www-data composer.json composer.lock ./
+RUN composer --no-interaction install --no-autoloader --no-scripts --no-suggest
+
+COPY --chown=www-data:www-data . ./
+
+ARG APP_ENV
```
giorgiosironi, line 31: these two seem to have default values (as images can be used in multiple environments and with different secrets configuration), so if that is the case the `ARG` themselves should have defaults

dashboard/src/Entity/Event.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @ORM\Column(type="guid")
+     * @ORM\Id
+     */
+    private $id;
+
+    /**
+     * @ORM\Column(type="string")
+     */
+    private $type;
```
giorgiosironi, line 34: It's probably worth it to add a basic storage and dump of the `data` field as it may be used to provide debugging information (no schema required)

seanwiseman:
I have been able to bring up the services, publish some event messages and see them coming through on the dashboard. 👍 

As it is understandably only processing events of type `article.published` (as far as I could see) maybe add the [example event](https://github.com/libero/walking-skeleton/pull/18#issuecomment-387322802) to the/a readme.

seanwiseman (approved):
I have nothing else to add at present, seems functional for its intended use. 👍

PR #19, opened 2018-05-04
Docker fixes

thewilkybarkid:
Fixes JS used in Journal's web profiler (/cc @diversemix), and sends PHP logs to stderr.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
             context: ./api-dummy
         volumes:
             - ./api-dummy:/srv/app
-            - /public/bundles
+            - /srv/app/public/bundles
             - /srv/app/var
             - /srv/app/vendor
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -26,7 +26,7 @@ services:
             context: ./journal
         volumes:
             - ./journal:/srv/app
-            - /public/bundles
+            - /srv/app/public/bundles
```
giorgiosironi, line 29: not sure what's inside here, but pointing out it will be persisted between updates to the code (being a volume)

giorgiosironi (approved):

docker-compose.yml
thewilkybarkid: Not much in comparison to `vendor`, but would definitely like to find a better solution.

PR #21, opened 2018-05-08
Start standardising RELAX NG

thewilkybarkid:
Looking at tidying up the sample schemas a bit, since they should be looked at closely (even though the actual content is important at the this, the mechanism is).

- Provide grouping
    - Also add some documentation
- Clear and consistent naming
- More extensible
    - Each element is in a `<define>` has two sub-`<define>`s, one for attributes, one for elements ('content'), even if they're empty (to allow for easy extension).

---

Used the following for inspiration:

- http://books.xmlschemata.org/relaxng/page2.html
- https://www.oasis-open.org/committees/download.php/51768/proposal-13112-rng-stage-3.html
- https://www.w3.org/MarkUp/2010/xhtml-m12n-relaxng-20101216/
- https://www.tei-c.org/release/xml/tei/custom/schema/relaxng/tei_all.rng

### DIFF
--- a/api-dummy/public/schemas/libero/article/body.rng
+++ b/api-dummy/public/schemas/libero/article/body.rng
@@ -20,7 +20,7 @@
         </define>
 
         <define name="libero.body.attributes">
-            <ref name="libero.attrib.lang"/>
+            <ref name="libero.attributes.lang"/>
         </define>
 
         <define name="libero.body.content">
--- a/api-dummy/public/schemas/libero/article/body.rng
+++ b/api-dummy/public/schemas/libero/article/body.rng
@@ -1,20 +1,32 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 
-<grammar xmlns="http://relaxng.org/ns/structure/1.0" ns="http://libero.pub">
+<grammar ns="http://libero.pub" xmlns="http://relaxng.org/ns/structure/1.0"
+    xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">
 
     <include href="../core.rng"/>
 
     <start>
-
-        <element name="body">
-            <ref name="libero.body.content"/>
-        </element>
-
+        <ref name="libero.body"/>
     </start>
 
-    <define name="libero.body.content">
-        <ref name="libero.attrib.lang"/>
-        <ref name="libero.blocks.full.model"/>
-    </define>
+    <div>
+        <a:documentation>body element</a:documentation>
+
+        <define name="libero.body">
+            <element name="body">
+                <ref name="libero.body.attributes"/>
+                <ref name="libero.body.content"/>
+            </element>
+        </define>
+
+        <define name="libero.body.attributes">
+            <ref name="libero.attrib.lang"/>
+        </define>
+
+        <define name="libero.body.content">
+            <ref name="libero.blocks.full.model"/>
+        </define>
+
+    </div>
 
 </grammar>
--- a/api-dummy/public/schemas/libero/article/front.rng
+++ b/api-dummy/public/schemas/libero/article/front.rng
@@ -1,26 +1,75 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 
-<grammar xmlns="http://relaxng.org/ns/structure/1.0" ns="http://libero.pub">
+<grammar ns="http://libero.pub" xmlns="http://relaxng.org/ns/structure/1.0"
+    xmlns:a="http://relaxng.org/ns/compatibility/annotations/1.0">
 
     <include href="../core.rng"/>
 
     <start>
-
-        <element name="front">
-            <ref name="libero.front.content"/>
-        </element>
-
+        <ref name="libero.front"/>
     </start>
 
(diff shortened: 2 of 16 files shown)


api-dummy/public/schemas/libero/extensions/youtube.rng
```diff
@@ -1,30 +1,85 @@
...
+
+        <define name="libero.blocks.youtube.content.id">
+            <element name="id">
+                <ref name="libero.blocks.youtube.content.id.attributes"/>
+                <ref name="libero.blocks.youtube.content.id.content"/>
+            </element>
+        </define>
+
+        <define name="libero.blocks.youtube.content.id.attributes">
+            <empty/>
```
thewilkybarkid, line 74: Example of something that the Libero part doesn't need (ie there's no attributes defined on this element), but leaves the easy and clear possibility of extension.

api-dummy/public/schemas/libero/extensions/youtube.rng
```diff
@@ -1,30 +1,85 @@
...
+        </define>
+
+        <define name="libero.blocks.youtube.content.id.content">
+            <data type="token">
```
thewilkybarkid, line 78: `token` seems preferable to `string` as it allows for whitespace before and after (which means we'd need to document whitespace expectations).

api-dummy/public/schemas/libero/article/body.rng
```diff
@@ -1,20 +1,32 @@
...
 
     <start>
+        <ref name="libero.body"/>
+    </start>
 
-        <element name="body">
-            <ref name="libero.body.content"/>
-        </element>
+    <div>
+        <a:documentation>body element</a:documentation>
```
thewilkybarkid, line 13: Alternatives:

- Comments
- XHTML

See http://books.xmlschemata.org/relaxng/relax-CHP-13-SECT-2.html.

api-dummy/public/schemas/libero/article/front.rng
```diff
@@ -1,26 +1,75 @@
...
+                <ref name="libero.front.content"/>
+            </element>
+        </define>
+
+        <define name="libero.front.attributes">
+            <ref name="libero.attrib.lang"/>
+        </define>
+
+        <define name="libero.front.content">
+            <interleave>
```
thewilkybarkid: This was missing, which meant a fixed ordering. This makes it flexible (since the order doesn't matter).

api-dummy/public/schemas/libero/extensions/youtube.rng
```diff
@@ -1,30 +1,85 @@
...
-    
-    <define name="libero.blocks.youtube.model">
-        <attribute name="height">
+
+    <div>
+        <a:documentation>youtube element</a:documentation>
+
+        <define name="libero.blocks.youtube">
+            <element name="youtube">
+                <ref name="libero.blocks.object.model"/>
```
thewilkybarkid, line 16: Something that hasn't been cleaned up yet.

thewilkybarkid:
This is incomplete, but is probably sufficient (since we're not making this production ready).

api-dummy/public/schemas/libero/article/body.rng
giorgiosironi: XHTML seems too specific, while this kind of tags allow for introspection by tools (whether it is a build or process validating a schema or something that displays the text of the annotations) in a way that comments probably won't

api-dummy/public/schemas/libero/article/body.rng
```diff
@@ -1,20 +1,32 @@
...
+                <ref name="libero.body.attributes"/>
+                <ref name="libero.body.content"/>
+            </element>
+        </define>
+
+        <define name="libero.body.attributes">
+            <ref name="libero.attributes.lang"/>
+        </define>
+
+        <define name="libero.body.content">
```
giorgiosironi, line 26: `attributes` is clear enough but is `content` a standard word?

giorgiosironi: don't want to overload it with, for example, the more specific `content` we use for various items like blog articles and such

api-dummy/public/schemas/libero/extensions/youtube.rng
```diff
@@ -1,30 +1,85 @@
...
+        <define name="libero.blocks.youtube.attributes.width.content">
             <data type="positiveInteger"/>
-        </attribute>
-        <element name="id">
-            <data type="string">
+        </define>
+
+    </div>
+
+    <div>
```
giorgiosironi, line 63: the tag you don't expect to show up, but its meaning is clear enough

giorgiosironi (approved):

api-dummy/public/schemas/libero/article/body.rng
thewilkybarkid: Borrowed from https://www.w3.org/MarkUp/2010/xhtml-m12n-relaxng-20101216/#s_relaxngrules_groups

seanwiseman (approved):

PR #22, opened 2018-05-08
Create Article store

seanwiseman:


article_store/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+version: '3'
```
thewilkybarkid: This will need to be moved to the main `docker-compose.yml` (assume it's just here temporarily).

seanwiseman: Yer just for local dev for this PR, then I will move it across.

article_store/articles/models.py
```diff
@@ -0,0 +1,22 @@
+import uuid
+
+from django.db import models
+
+
+PUBLISHED = 'Published'
+UNPUBLISHED = 'Unpublished'
```
thewilkybarkid: Might want to think about this being configurable (eg you might want 'preview', 'ready' and 'published').

article_store/articles/models.py
```diff
@@ -0,0 +1,22 @@
...
+
+ARTICLE_STATUSES = (
```
thewilkybarkid: Might want to think about this being configurable (eg you might want 'preview', 'ready' and 'published').

thewilkybarkid: Though might be ok to ignore for the skeleton.

article_store/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.6.4-jessie
+
+RUN pip install pipenv
+
+ADD . /srv/app/
+WORKDIR /srv/app/
+
+RUN pipenv install --system
```
giorgiosironi: does this mean system-wide as in overwriting the Python packages provided by the container's distribution? I'd be more cautious to avoid interaction between non-project files and project files, if it's just a default choice?

article_store/articles/tests/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+
+@pytest.fixture
+def content_xml_data() -> str:
+    return '''
+    <root>
+        <content-list>
+            <list-item>
+                <language>en</language>
+                <name>front</name>
+                <model>http://localhost:8081/schemas/scielo/article/body.rng</model>
```
thewilkybarkid: There's a question here of do we allow multiple schemas, or do you have to run an article store (should be something like 'content store', but that sounds even worse!) per schema? (eg could you have scholarly articles and blog posts in the same instance, or would they have to be separate)

Either way, this value should be configuration on the app rather than input.

article_store/articles/tests/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+@pytest.fixture
+def content_xml_data() -> str:
+    return '''
+    <root>
+        <content-list>
+            <list-item>
+                <language>en</language>
+                <name>front</name>
```
thewilkybarkid: These two should be extracted from the data (root element name, and the required `xml:lang` attribute), but might as well leave for now.

article_store/articles/tests/conftest.py
```diff
@@ -0,0 +1,134 @@
...
+                                  model='http://localhost:8081/schemas/scielo/article/front.rng',
+                                  content=article_0065_pt_front_payload_xml)
+
+
+@pytest.fixture
+def content_xml_data() -> str:
+    return '''
+    <root>
+        <content-list>
+            <list-item>
```
thewilkybarkid: These three levels should be replaced with something like `<article><parts>`, but we'll sort it out later.

article_store/articles/api.py
```diff
@@ -0,0 +1,204 @@
...
+        article, created = Article.objects.get_or_create(id=article_id)
+        article_version = ArticleVersion.objects.create(version=article.next_version,
+                                                        article=article)
+
+        for content_item in request.data.get('content-list', []):
+            Content.objects.create(article_version=article_version, **content_item)
+
+        serializer = ArticleVersionSerializer(article_version)
+
+        return Response(serializer.data, status=status.HTTP_201_CREATED)
```
thewilkybarkid: `201` is fine given we don't have Airflow yet, but will have to become `202`.

article_store/articles/utils.py
```diff
@@ -0,0 +1,26 @@
...
+    'en'
+
+    >>> parse_accept_language_header('es')
+    'es'
+
+    :param lang_str: str
+    :return: str
+    """
+    try:
+        return parse_accept_lang_header(lang_str)[-1][0]
```
thewilkybarkid: Fine for now as Journal only requests one language, but we'll need to follow the HTTP spec in the future.

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+PREVIEW = 'preview'
+PUBLISHED = 'published'
+READY = 'ready'
+
+ARTICLE_STATUSES = (
+    (PREVIEW, 'preview'),
+    (PUBLISHED, 'Published'),
+    (READY, 'ready'),
```
thewilkybarkid: Ok for now, will will probably have to be configurable.

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __str__(self):
+        return f'Article: {self.article_id}, Version {self.version}'
+
+
+class Content(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    article_version = models.ForeignKey(ArticleVersion, related_name='content_items', on_delete=models.CASCADE)
+    language = models.CharField(max_length=50)
+    name = models.CharField(max_length=250)  # TODO has to be unique per version ??
```
thewilkybarkid: Per version per language?

article_store/core/urls.py
```diff
@@ -0,0 +1,15 @@
+from django.conf import settings
+from django.conf.urls.static import static
+from django.contrib import admin
+from django.urls import path, include
+
+
+urlpatterns = [
+    path('admin/', admin.site.urls),
+    path('articles/', include('articles.urls')),
```
thewilkybarkid: This might need to be part of deployment, so you could have `/blog-articles` or whatever instead (related to https://github.com/libero/walking-skeleton/pull/22/files#r188602069).

article_store/articles/api.py
```diff
@@ -0,0 +1,204 @@
...
+            return HttpResponse(payload, status=status.HTTP_200_OK, content_type="application/xml")
+
+        except ValidationError as err:
+            return Response({'error': f'Invalid article ID - {err}'}, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+        except AttributeError as err:
+            return Response({'error': f'Content does not exist - {err}'}, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+        except ObjectDoesNotExist as err:
+            return Response({'error': f'{err}'}, status=status.HTTP_406_NOT_ACCEPTABLE)
```
thewilkybarkid: We'll want to use the XML version of RFC 7807, but ignore for now.

article_store/articles/tests/conftest.py
seanwiseman: Agreed. With a little more time on the parser, it can easily be done.

article_store/articles/tests/conftest.py
seanwiseman: Sure. You can actually call the top level element what you like now as it's only the container and is thrown away and the others are just defaults that we can change.

article_store/Dockerfile
seanwiseman: I was just following on from a few example images such as [this](https://hub.docker.com/r/kennethreitz/pipenv/~/dockerfile/). I don't really have a preference or have had enough `docker` + `pipenv` experience yet. Happy to work with whatever is decided.

article_store/articles/models.py
seanwiseman: OK, would only take some validation pre-save to achieve. Can probably wait as these are eventually going to be pulled from the XML data.

seanwiseman:
I have added a basic [README](https://github.com/libero/walking-skeleton/pull/22/commits/ecf9f99dfb5bdc52b69fe0fd2422142f9a67735b#diff-744109e63d23ae52d86a3293fe6c336a) detailing the endpoints available. Should be enough to post some article data in. With the current root `docker-compose.yml` the `journal` is point to the `article-store` so you should see the article data appearing there.

article_store/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+Deletes an entire article including all versions and content.
+
+---------------------
+
+`POST` `PUT` `DELETE`
+
+`/articles/<article_id>/<version>`
+
+`POST` Create an article version, can include content items.
```
giorgiosironi: an example `curl` would be useful here, I tried:
```
curl -v -d '<root>...' http://localhost:8000/articles/0065/1 -X POST
```
but got
```
Exception Type: TypeError at /articles/0065/1
Exception Value: post() got an unexpected keyword argument 'version'
```

docker-compose.yml
```diff
@@ -21,6 +21,30 @@ services:
...
+    postgres:
+        image: postgres:9.6.8-alpine
+        environment:
+          - POSTGRES_USER=root
+          - POSTGRES_PASSWORD=rootpassword
+          - POSTGRES_DB=default
+        ports:
+          - '5432:5432'
+        volumes:
+          - 'postgres:/var/lib/postgresql/data'
```
giorgiosironi: Stopping `docker-compose up` and restarting it doesn't reproduce data loss.
`docker-compose down` does remove the data, even without `-v`; ~I think this is an expected behavior of `docker-compose down` for certain kinds of volumes.~ should not happen and in fact I see volumes surviving with `docker volume ls`, but perhaps there is data outside of them.

giorgiosironi: There seems to be no data in the container itself:
```
$ docker diff walkingskeleton_postgres_1
C /run/postgresql
```

giorgiosironi: A new `docker-compose up` shows some state (last shutdown) is being kept by its volume for the `postgres` container:
```
$ docker-compose up
Creating walkingskeleton_article-store_1   ... done
Creating walkingskeleton_api-dummy-nginx_1 ... done
Creating walkingskeleton_article-store_1   ... 
Creating walkingskeleton_journal-fpm_1     ... done
Creating walkingskeleton_journal-fpm_1     ... 
Creating walkingskeleton_journal-nginx_1   ... done
Attaching to walkingskeleton_postgres_1, walkingskeleton_api-dummy-fpm_1, walkingskeleton_article-store_1, walkingskeleton_api-dummy-nginx_
1, walkingskeleton_journal-fpm_1, walkingskeleton_journal-nginx_1
postgres_1         | LOG:  database system was shut down at 2018-05-16 16:07:53 UTC
postgres_1         | LOG:  MultiXact member wraparound protections are now enabled
postgres_1         | LOG:  database system is ready to accept connections
postgres_1         | LOG:  autovacuum launcher started
```
but still after this happens, migrations are run again:
```
article-store_1    | Operations to perform:
article-store_1    |   Apply all migrations: admin, articles, auth, contenttypes, sessions
article-store_1    | Running migrations:
article-store_1    |   Applying contenttypes.0001_initial... OK
article-store_1    |   Applying auth.0001_initial... OK
article-store_1    |   Applying admin.0001_initial... OK
article-store_1    |   Applying admin.0002_logentry_remove_auto_add... OK
...
```

giorgiosironi: Database seems empty, maybe article store is writing somewhere else like an in-memory db?
```
default=# use postgres
default-# \dt
No relations found.
default-# use default
default-# \dt
No relations found.
```

giorgiosironi: `app.cfg` in the container seems correct:
```
[database]
name: default
engine: django.db.backends.postgresql
;host = localhost
host = postgres
port = 5432
password = rootpassword
user = root
```

giorgiosironi: I think this area is misfiring: https://github.com/libero/walking-skeleton/pull/22/files#diff-605ec7990a5cf92de1391cd93b7b7e99R71

article_store/Dockerfile
giorgiosironi: Tried without `--system` in another PR and it seems to work correctly, I'd go for that as the conservative option

article_store/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+WORKDIR /srv/app/
+
+CMD python manage.py migrate && python manage.py runserver 0.0.0.0:8000
```
giorgiosironi: (with that in mind, this becomes `pipenv run python ...`

article_store/README.md
```diff
@@ -0,0 +1,85 @@
+# Article Store
+
+An XML data store for the Libero Walking skeleton.
+
+## Endpoints
+`GET`
```
giorgiosironi: was the POST on `/articles` removed?

article_store/articles/fixtures/0065_en_front.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" ?>
```
giorgiosironi: what are fixtures used for? Perhaps they can be kept with the tests?

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+from django.db import models
+
+
+PREVIEW = 'preview'
+PUBLISHED = 'published'
+READY = 'ready'
+
+ARTICLE_STATUSES = (
+    (PREVIEW, 'preview'),
+    (PUBLISHED, 'Published'),
```
giorgiosironi: inconsistent uppercase letter

article_store/README.md
seanwiseman: OK will put some example urls along with the data in the examples to clarify this 👍.

You can't specify a version in the POST as it just creates the next version for you based on existing versions. You can specify the version using PUT (if it exists) as this will modify an existing version.

docker-compose.yml
seanwiseman: You are correct.

I have now changed the`CONF.get` calls to [`CONF.getboolean`](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.getboolean) where necessary.

Thank you 👍.

article_store/README.md
seanwiseman: As it seemed more logical that the users should decide the article ids (especially if porting existing content), I removed POST from this endpoint (which auto-generated a random article id).

Now you have to specify one using `/articles/<article_id>` obviously has to be unique still.

article_store/articles/fixtures/0065_en_front.xml
seanwiseman: Yes just for the test suite.

I have now nested them in the tests directory.

article_store/Dockerfile
seanwiseman: OK Done 👍

seanwiseman:
Will leave event publisher until the next PR, it's not a lot of work but keeps this one smaller and focused on data in and out.

giorgiosironi:
I get this error:
```
$ cat sample_version.xml 
<root>
    <content-list>
        <list-item>
            <language>en</language>
            <name>front</name>
            <model>http://localhost:8081/schemas/scielo/article/front.rng</model>
            <content>
                <front xmlns="http://libero.pub" xmlns:scielo="http://scielo.org" xml:lang="en">
                    <id>0065-1737-azm-31-03-0367</id>
                    <title>FOOBAR 9999 (Lepidoptera: Tortricidae) (Heinrich),
                        structural parameters and natural regeneration </title>
                    <abstract id="abstract">
                        <p>Foo</p>
                    </abstract>
                </front>
            </content>
        </list-item>
        <list-item>
            <language>en</language>
            <name>body</name>
            <model>http://localhost:8081/schemas/scielo/article/front.rng</model>
            <content>
                <body xmlns="http://libero.pub" xmlns:scielo="http://scielo.org" xml:lang="en">
                    <p>Foo</p>
                </body>
            </content>
        </list-item>
    </content-list>
</root>
```
```
$ curl -v -H "Content-Type: application/xml" http://localhost:8000/articles
/0064 -X POST -d "@./sample_version.xml"
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> POST /articles/0064 HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/7.47.0
> Accept: */*
> Content-Type: application/xml
> Content-Length: 1091
> Expect: 100-continue
> 
< HTTP/1.1 100 Continue
* We are completely uploaded and fine
< HTTP/1.1 201 Created
< Date: Thu, 17 May 2018 09:25:43 GMT
< Server: WSGIServer/0.2 CPython/3.6.4
< Content-Type: application/xml; charset=utf-8
< Allow: GET, POST, PUT, DELETE, HEAD, OPTIONS
< X-Frame-Options: SAMEORIGIN
< Content-Length: 1184
< Vary: Cookie
< 
<?xml version="1.0" encoding="utf-8"?>
<root><id>67cc5011-dece-4213-b192-a1b4ac6fccf5</id>
...
```
http://localhost:8080/ shows a list of one article, correctly.
http://localhost:8080/articles/0065-1737-azm-31-03-0367 then explodes:
```
Client error: `GET http://article-store:8000/articles/0065-1737-azm-31-03-0367/latest/front` resulted in a `406 Not Acceptable` response:
<?xml version="1.0" encoding="utf-8"?>
<root><error>Article matching query does not exist.</error></root>
```
Maybe it's the id in the POST URL not matching?

article_store/README.md
giorgiosironi: :+1:

article_store/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+Example:
+
+`curl -v http://localhost:8000/articles/0065 -X DELETE`
+
+---------------------
+
+`POST`
+
+`/articles/<article_id>`
```
giorgiosironi: I think this should be `/articles/<article_id>/versions` as otherwise it's not clear what `POST` is creating (could be another characteristic of an article)

giorgiosironi:
Yes, `curl -v -H "Content-Type: application/xml" http://localhost:8000/articles/0065-1737-azm-31-03-0367 -X POST -d "@./sample_version.xml"` works instead. Not sure if it can be easily checked in the POST implementation but the README example should be updated with the corresponding id in the URL.

article_store/README.md
giorgiosironi: In general that applies to various URLs later on like the `PUT` or the `DELETE` of a version. It also leaves the space for `GET /articles/<article_id>/versions` to return some kind of lists of versions, that I can't be obtain at the moment.

article_store/articles/urls.py
```diff
@@ -0,0 +1,22 @@
...
+router_v1.register('articles', ArticleViewSet, base_name='article')
+
+
+urlpatterns = [
+    path('/api/v1/', include((router_v1.urls, 'api_v1'))),
+    path('/<str:article_id>', ArticleXMLAPIView.as_view(), name='article-xml'),
+    path('/<str:article_id>/<str:version>', ArticleXMLAPIView.as_view(), name='article-version'),
+    path('/<str:article_id>/<str:version>/<str:part>', ArticleXMLAPIView.as_view(), name='article-content'),
+    path('', ArticleListAPIView.as_view(), name='article-list'),
+]
```
giorgiosironi: I get some warnings:
```
article-store_1    | WARNINGS:
article-store_1    | ?: (urls.W002) Your URL pattern '/<str:article_id>' [name='article-xml'] has a route beginning with a '/'. Remove this slash as it is unnecessary. If this pattern is targeted in an include(), ensure the include() pattern has a trailing '/'.
article-store_1    | ?: (urls.W002) Your URL pattern '/<str:article_id>/<str:version>' [name='article-version'] has a route beginning with a '/'. Remove this slash as it is unnecessary. If this pattern is targeted in an include(), ensure the include() pattern has a trailing '/'
```

article_store/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.6.4-jessie
+
+RUN pip install pipenv
+
+ADD . /srv/app/
+WORKDIR /srv/app/
+
+RUN pipenv install --deploy
```
giorgiosironi: :+1:

article_store/Pipfile
```diff
@@ -0,0 +1,28 @@
...
+verify_ssl = true
+name = "pypi"
+
+
+[packages]
+
+django = "~=2.0.5"
+django-debug-toolbar = "~=1.9.1"
+djangorestframework = "~=3.8.2"
+"psycopg2-binary" = "*"
```
giorgiosironi: why this does package name have double quotes and the others haven't? :thinking:

article_store/app.cfg
```diff
@@ -0,0 +1,19 @@
+[django]
+ci = false
+debug = true
+allowed_hosts = web,article-store,localhost,127.0.0.1
```
giorgiosironi: `web` is?

article_store/app.cfg
```diff
@@ -0,0 +1,19 @@
...
+secret_key = FAKE-KEY-DO-NOT-USE-IN-PRODUCTION
+
+[database]
+name: default
+engine: django.db.backends.postgresql
+;host = localhost
+host = postgres
```
giorgiosironi: This seems to be the actual configuration file (pointing to `postgres:5432`) therefore should not be in the image, but mounted from the outside as a volume by docker-compose.

The alternative would be to pass in environment variables, but the values here seem involved enough that we don't want to define one for each.

article_store/articles/admin.py
```diff
@@ -0,0 +1,60 @@
...
+        'article_version',
+        'language',
+        'name',
+    )
+    list_display_links = ['id']
+
+
+admin.site.register(Article, ArticleAdmin)
+admin.site.register(ArticleVersion, ArticleVersionAdmin)
+admin.site.register(Content, ContentAdmin)
```
giorgiosironi: who can access the admin UI and how?

article_store/articles/urls.py
seanwiseman: These are because of the removal of the trailing `/` (which Django likes) on the `/articles` endpoint. Which means I have to include it in subsequent path matches.

I will change a setting which should remove this check as its not needed.

article_store/articles/api.py
```diff
@@ -0,0 +1,206 @@
...
+
+        if not article_id or not Article.id_is_valid(article_id):
+            return Response({'error': 'Please provide a valid article id'}, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+        article, created = Article.objects.get_or_create(id=article_id)
+        article_version = ArticleVersion.objects.create(version=article.next_version,
+                                                        article=article)
+
+        for content_item in request.data.get('content-list', []):
+            Content.objects.create(article_version=article_version, **content_item)
```
giorgiosironi: are there transaction boundaries here? I think at least the `ArticleVersion` and the `Content` objects should be inside the same database transaction

article_store/articles/api.py
```diff
@@ -0,0 +1,206 @@
...
+        :param request:
+        :param article_id:
+        :param version:
+        :return: class: `HttpResponse`
+        """
+        if article_id and version:
+            article_version = ArticleVersion.objects.get(version=version, article_id=article_id)
+
+            old_content = Content.objects.filter(article_version=article_version)
+            old_content.delete()
```
giorgiosironi: I think it's fine to delete and re-add, but also by ensuring it's a single database transaction so that clients never see an incomplete version

article_store/articles/api.py
```diff
@@ -0,0 +1,206 @@
...
+            if not version:
+                article.delete()
+            else:
+                if version == self.latest:
+                    version = article.latest_version
+
+                article_versions = ArticleVersion.objects.filter(article_id=article_id)
+
+                for article_version in article_versions:
+                    if article_version.version >= int(version):
```
giorgiosironi: if I delete version 2, then version 3 is deleted as well: no opinion on that, but it leaves the user freedom. At eLife we practically never delete anything, only unpublished versions ingested by error.

giorgiosironi: (freedom as opposed to: refusing to delete version 2 if a version 3 exists)

article_store/articles/api.py
```diff
@@ -0,0 +1,206 @@
...
+    root = Element('{%s}articles' % name_space['libero'], nsmap=name_space)
+
+    for article in articles:
+        child = SubElement(root, '{%s}article' % name_space['libero'])
+        child.text = str(article.id)
+
+    return root
+
+
+class ArticleXMLAPIView(APIView):
```
giorgiosironi: this is called `ArticleXML...` but comparing it to `ArticleList...` (which also uses XML) its difference is being oriented to a single article and its components rather than a list. Can be called `ArticleItem` as opposed to `List`, or `ArticleAggregate...` for the DDD-style?
(`...` stands for `APIView`, which is the superclass)

article_store/Pipfile
seanwiseman: No idea was generated by pipenv... 🤔 ...will remove quotes.

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+
+from django.core.validators import RegexValidator
+from django.db import models
+
+
+PREVIEW = 'preview'
+PUBLISHED = 'published'
+READY = 'ready'
+
+ARTICLE_STATUSES = (
```
giorgiosironi: these are `ARTICLE_VERSION_STATUSES`, right?

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+    (PREVIEW, 'preview'),
+    (PUBLISHED, 'published'),
+    (READY, 'ready'),
+)
+
+ARTICLE_ID_FORMAT = r'^[A-Za-z0-9\-._]+$'
```
giorgiosironi: if this is not wide-enough, it can be expanded easily later e.g. adding `\+`, as long as it's URL-friendly

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+class Article(models.Model):
+    id = models.CharField(primary_key=True, max_length=250,
```
giorgiosironi: I agree this should be a string for maximum compatibility (we used integers up to now)

article_store/app.cfg
seanwiseman: ...the old name for the container and no longer needed 👍 will remove

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+                          validators=[RegexValidator(regex=ARTICLE_ID_FORMAT)])
+
+    article_id_format = ARTICLE_ID_FORMAT
+    default_version = 0
+
+    def __str__(self):
+        return f'{self.id}'
+
+    @property
+    def latest_version(self) -> int:
```
giorgiosironi: seeing how this is used, perhaps it (and `self.default_version`) could be called `version_count` instead?
Because an empty article as `0` versions when you count them, but `0` is not a valid version instead. There are no holes like a missing version 2 (according to the DELETE API behavior) so they should be equivalent.

article_store/articles/models.py
seanwiseman: I don't mind, I was going by [this](https://github.com/libero/walking-skeleton/blob/api/schema/articles.yaml#L34).

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+        """Check if a `str` value is valid against `article_id_format`
+
+        :param article_id: str
+        :return: bool
+        """
+        return True if re.match(cls.article_id_format, article_id) else False
+
+
+class ArticleVersion(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
giorgiosironi: UUID for `ArticleVersion` may be an overkill because there is a strict order of versions and they must be generated over a single-master, linearizable database (`postgres`).

Since `article`+`version` is a composite key I agree a generated primary key makes things easier especially for the ORM. Could just be an incremental id? It's never going to be exposed to the API.

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+PREVIEW = 'preview'
+PUBLISHED = 'published'
+READY = 'ready'
+
+ARTICLE_STATUSES = (
+    (PREVIEW, 'preview'),
+    (PUBLISHED, 'published'),
+    (READY, 'ready'),
```
giorgiosironi: not sure what is `ready` (the other two are clear)?

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+        return True if re.match(cls.article_id_format, article_id) else False
+
+
+class ArticleVersion(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    article = models.ForeignKey(Article, related_name='versions', on_delete=models.CASCADE)
+    status = models.CharField(max_length=50, choices=ARTICLE_STATUSES, default=PREVIEW)
+    version = models.IntegerField(default=1)
+
+    # TODO needs check on version number unique, e.g. can't save x2 version 1's
```
giorgiosironi: ...for the same article; composite unique index on `article`, `version`

article_store/articles/models.py
seanwiseman: Yes, you're right, it has now evolved into `ARTICLE_VERSION_STATUSES`.

article_store/articles/models.py
seanwiseman: Was advised [here](https://github.com/libero/walking-skeleton/pull/22#discussion_r186752381), not sure I could give you the exact use case for it 🤔, maybe a need to know when an articles state is between `preview` > .... > `published`

article_store/articles/models.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __str__(self):
+        return f'Article: {self.article_id}, Version {self.version}'
+
+
+class Content(models.Model):
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
```
giorgiosironi: I think a UUID for `Content` is fine because there is no clear ordering to maintain between, it could allow `Content` objects ids to be generated elsewhere and pushed in without conflict. For events, I went for UUID1 (time-based) rather than 4 (random-based) as it encodes some debugging information about when it is created, and is more difficult for implementations to mess up since the time-based prefix can't collide.

article_store/articles/models.py
giorgiosironi: Even if they are pulled from the XML data, if may be helpful to always include them as table columns for ease of querying/debugging/aggregating/indexing

article_store/articles/xml_parser.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def register_namespaces(name_spaces: Tuple[Tuple[str, str]]) -> None:
+    """Register `n` number of xml namespaces
+
+    :param name_spaces: Tuple
+    :return:
+    """
+    for name_space in name_spaces:
+        register_namespace(*name_space)
```
giorgiosironi: it's a pity that this is global rather than being on a `parser` object, but can't change a standard library :man_shrugging:

article_store/core/settings.py
```diff
@@ -0,0 +1,144 @@
...
+
+MIDDLEWARE = [
+    'django.middleware.security.SecurityMiddleware',
+    'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.common.CommonMiddleware',
+    'django.middleware.csrf.CsrfViewMiddleware',
+    'debug_toolbar.middleware.DebugToolbarMiddleware',
+    'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'django.contrib.messages.middleware.MessageMiddleware',
+    'django.middleware.clickjacking.XFrameOptionsMiddleware',
```
giorgiosironi: is this something related to the admin UI? Seems a API wouldn't need to worry about this?

docker-compose.yml
```diff
@@ -21,6 +21,30 @@ services:
             - ./api-dummy/public:/srv/app/public
         depends_on:
             - api-dummy-fpm
+    article-store:
+        build:
+            context: ./article_store
+        command: >
+                /bin/bash -c "
+                    until echo > /dev/tcp/postgres/5432; do sleep 1; done
+                    pipenv run python manage.py migrate articles
```
giorgiosironi: why two separate `migrate` commands? As a Django newbie here

docker-compose.yml
```diff
@@ -21,6 +21,30 @@ services:
...
+                    pipenv run python manage.py migrate
+                    pipenv run python manage.py runserver 0.0.0.0:8000
+                "
+        ports:
+          - '8000:8000'
+        depends_on:
+          - postgres
+    postgres:
```
giorgiosironi: agreed that this should be prefixed with `article-store`

docker-compose.yml
```diff
@@ -31,7 +55,7 @@ services:
             - /srv/app/vendor
         environment:
             - APP_ENV=dev
-            - APP_LIBERO_API=http://api-dummy-nginx
+            - APP_LIBERO_API=http://article-store:8000
```
giorgiosironi: if you merge `master` into this branch, this configuration will end up in https://github.com/libero/walking-skeleton/blob/master/docker-compose.yml#L32 as there's a gateway now

article_store/articles/models.py
giorgiosironi: Statuses in general can be ignored for now, given the API creates an immediately published article version

article_store/articles/admin.py
seanwiseman: Only the cool kids 😎 ... I will put a section in the readme.

It's not essential but is nice to look through the data, especially at this early stage.

docker-compose.yml
seanwiseman: It's actually an artefact of initial work, its no longer required. Will reduce to just `migrate`.

article_store/articles/models.py
seanwiseman: Will change to incremental id for simplicity then.

seanwiseman: It's not going to allow a nice migration from `uuid` to `int` for existing rows, you can manually script a conversion but it's not worth it considering the minimal data we would have put in these instances:

`cannot cast type uuid to integer`

I will reset the migrations to wipe the `uuid` column, will just mean you will have to start clean.

article_store/articles/models.py
seanwiseman: Changed to `UUID1` for project consistency.

article_store/articles/api.py
seanwiseman: So... `ArticleXMLAPIView` -> `ArticleAggregateAPIView` ?

article_store/articles/models.py
seanwiseman: This came out of having to handle the `latest` keyword as a possible version value in a URL, just followed the naming from there.

I don't mind either way, I have refactored to `count` based solution.

seanwiseman:
I think I have addressed all of today's feedback 😅, thank you @giorgiosironi  and @thewilkybarkid .

docker-compose.yml
```diff
@@ -37,23 +37,24 @@ services:
             context: ./article-store
         volumes:
             - ./article-store:/srv/app
+        environment:
+            - ALLOWED_HOSTS=article-store
+            - APP_ENV=dev
+            - APP_SECRET=f6a87c2e031ecadc0d7eb442729435c8
+            - DATABASE_URL=postgres://postgres:@article-store-postgres:5432/postgres
```
seanwiseman: 👍

article-store/articles/api.py
```diff
@@ -0,0 +1,209 @@
...
+    root = Element('{%s}articles' % name_space['libero'], nsmap=name_space)
+
+    for article in articles:
+        child = SubElement(root, '{%s}article' % name_space['libero'])
+        child.text = str(article.id)
+
+    return root
+
+
+class ArticleXMLAPIView(APIView):
```
giorgiosironi: `ArticleItemAPIView` would be the simplest and clear name (it's a class name not exported to other project so easy to rename at any time)

article-store/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN mkdir -p /srv/app && \
+    chown www-data:www-data /srv/app
+
+USER www-data
+WORKDIR /srv/app
+
+COPY --chown=www-data:www-data Pipfile Pipfile.lock ./
+RUN pipenv install --deploy
+
+COPY --chown=www-data:www-data . ./
```
giorgiosironi, line 17: Usually the pattern for webserver (and by extension application server) user is that they run as `www-data` and all files outside of `/var/something` (or `var/something`) are owned by someone else (here it could be just `root`). That way compromised (Python) code still can't modify the code of the application. Therefore I'd strip the `--chown` in this file and push `USER www-data` down to the `CMD`.

article-store/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+Example:
+
+`curl -v http://localhost:8000/articles/0065 -X DELETE`
+
+---------------------
+
+`POST`
+
+`/articles/<article_id>`
```
giorgiosironi: having a sub-resource `/articles/<article_id>/versions` to house version manipulation needs further discussion, as it follows the REST pattern of POSTing on a collection rather than on an item, and it's a published API

docker-compose.yml
giorgiosironi: I see `.cfg` files have vanished, so I assume all configuration is performed here :+1:

giorgiosironi (commented):
Second round is in, mostly about the published API

article-store/README.md
seanwiseman: Agreed. Once the decision has been made this can easily be implemented.

article-store/README.md
```diff
@@ -0,0 +1,130 @@
...
+            </content>
+        </list-item>
+    </content-list>
+</root>
+```
+
+---------------------
+`PUT` `DELETE`
+
+`/articles/<article_id>/<version>`
```
thewilkybarkid: These need updating.

article-store/Dockerfile
thewilkybarkid: Agree on this, but the other apps aren't perfect either (since this is just throwaway) so don't mind whether this is done or not.

seanwiseman: Agree, though happy to leave for the time being.

thewilkybarkid (approved):
:shipit:

PR #23, opened 2018-05-09
Add API schema

thewilkybarkid:


thewilkybarkid:
Running into problems with OpenAPI again: not being able to nest paths makes it hard to compose schemas, and it's still _very_ JSON focused (so can't reference RELAX NG schemas).

api-doc/api.raml
```diff
@@ -0,0 +1,18 @@
...
+title: eLife's Libero walking skeleton
+
+baseUri: http://localhost:8085
+
+resourceTypes:
+    articles: !include articles.raml
+
+types:
+    ArticleVersionFront:
+        type: !include http://api-dummy-nginx:80/schemas/elife/article/front.rng # don't work as contains localhost references
```
thewilkybarkid: Wouldn't normally be a problem, but for now `localhost:8086` will need to be proxied to `schemas:80` somehow.

giorgiosironi: are these references meant to be moved to `schemas:80`? `api-dummy-nginx:80` seems to have the same problems

giorgiosironi: I see the `Can not resolve http://api-dummy-nginx:80/articles/00666/latest/front` now

giorgiosironi (commented):
General comments:

- the level of introspection around paths, HTTP methods, media types is adequate for documentation purposes
- once we get to an XML response, samples can be as useful schemas; it seems they have some problems but are supported in the interface
- navigation to ` <include>` related .rng files would be useful to explore a schema in depth (I copy and paste the http://localhost:8086/extensions/abstract.rng URLs to the location bar)
- validation is an end goal that would add a lot of value; if we had a Docker container that can be run on the cli or an API to call to validate one of the XML responses, it would open up testing possibilities for disparate services like the article store

api-dummy/public/schemas/libero/vendor/csl/csl-data.rng
```diff
@@ -1,255 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
thewilkybarkid, line 1: These files were left over from an experiment and shouldn't have been committed.

thewilkybarkid:
This is probably ok for now. There's some things still to think about:

1. **Presentation.** Libraries for both OpenAPI and RAML aren't great. Might have to produce our own (😞).
2. **Validation.** How to validate implementing services.
3. **Where the schemas should live.** Should the Libero article store implement an API specification, or define it itself?
4. **Is this needed?** Is this any better than having static documentation?

giorgiosironi:
> Presentation. Libraries for both OpenAPI and RAML aren't great. Might have to produce our own (disappointed).

I can understand the need, probably needs to be made scalable to many different content types or schemas.

> Validation. How to validate implementing services.

My summary is that I wasn't a believer in static validation of data, but the `api-raml` experience opened my eyes to complex/large content types to the point that I think it's necessary to keep documentation up-to-date. The generated documentation can be obviated by lots of examples, but the alternative to a schema is integration testing, which works on a subset of examples but isn't viable for all the intricate kinds of legacy content and combinations of data types in an article.

> Where the schemas should live. Should the Libero article store implement an API specification, or define it itself?

In general we have seen multiple services relying on the same content types (e.g. search, recommendations), so that pushes for its separation. Also, allows more control of the evolution rather than having each service implement its own versioning (easier to switch a number from `1.1` to `1.2` than to upgrade a library).

> Is this needed? Is this any better than having static documentation?

Personally I haven't used much the presentation part of `api-raml`; usually going directly to the source as it maps in a simple way to abstractions; for example, I have to fix or add a `figure` block, easy to see which are the mandatory fields in https://github.com/elifesciences/api-raml/blob/develop/src/blocks/figure.v2.yaml#L38 although the juggling of multiple files linking to each other makes it more difficult. 

Experimenting with this to see if it allows that use case, which is more peculiar to me ("integrate and debug multiple services") than development of a service in isolation. It didn't help that in some cases the `api-raml` version to use was a branch or a pull request rather than pointing your browser to https://api.elifesciences.org/ which came later.

giorgiosironi:
The example in the `Response` of http://localhost:8085/#articles__id__versions__version__front_get still has some wiring problem:
```
Can not resolve http://api-dummy-nginx:80/articles/00666/versions/latest/front
```
May need some forwarding through the gateway?

giorgiosironi:
The links like
http://localhost:8086/extensions/doi.rng
seem to go to Airflow, so the port needs to be changed?

api-doc/api.raml
```diff
@@ -0,0 +1,21 @@
...
+    Example of a composed Libero API schema.
+
+baseUri: http://localhost:8085
+
+resourceTypes:
+    articles: !include articles.raml
+
+types:
+    ArticleVersionFront:
+        type: !include http://api-dummy-nginx:80/schemas/elife/article/front.rng
```
thewilkybarkid: Reading the RAML spec, I think this is technically invalid. Seems ok to process JSON Schema or XML Schema, but doesn't allow for other things. raml2html emits a warning (actually in raml2obj) and just outputs the RELAX NG instead.

thewilkybarkid: Related: https://github.com/raml-org/raml-spec/issues/117

api-doc/api.raml
```diff
@@ -0,0 +1,21 @@
...
+description: |
+    Example of a composed Libero API schema.
+
+baseUri: http://localhost:8085
+
+resourceTypes:
+    articles: !include articles.raml
+
+types:
+    ArticleVersionFront:
```
thewilkybarkid: This might need to be rethought. In the event we want to run the article store multiple times (once for scholarly articles, once for blog posts), this wouldn't work? (Thought are we going to allow the same service to be run multiple times?)

thewilkybarkid:
> My summary is that I wasn't a believer in static validation of data, but the api-raml experience opened my eyes to complex/large content types to the point that I think it's necessary to keep documentation up-to-date. The generated documentation can be obviated by lots of examples, but the alternative to a schema is integration testing, which works on a subset of examples but isn't viable for all the intricate kinds of legacy content and combinations of data types in an article.

Sounds like JSON Schema was directly useful as opposed to the RAML?

> In general we have seen multiple services relying on the same content types (e.g. search, recommendations), so that pushes for its separation. Also, allows more control of the evolution rather than having each service implement its own versioning (easier to switch a number from 1.1 to 1.2 than to upgrade a library).

That refers to snippets, which we aren't looking at replicating, so they're naturally separated. My thought is about contract and implementation: we deliberately separated them, but it adds to complexity here. It did allow for the API Dummy, for example, but that's going to be hard for Libero given we don't know what your API will be?

giorgiosironi:
> Sounds like JSON Schema was directly useful as opposed to the RAML?

Sort of, the RAML paths were simple enough (`/articles/:id`) that they were easier to get right without reading a formal definition; `/search` was an example of one that had many (5-6) query parameters and hence benefitted from defining them, their types and whether they were optional.

giorgiosironi:
>  contract and implementation: we deliberately separated them, but it adds to complexity here. It did allow for the API Dummy, for example, but that's going to be hard for Libero given we don't know what your API will be?

We also don't know what the implementation will be (as in allowing for other implementations of an article store or other components) so as long as there are multiple possibilities for that a separate contract can be reused by all of them.

thewilkybarkid:
I think I've been able to reasonably solve the composition question with regards to using the same service twice. I've added examples of how you can extend it too, so if someone implemented the article store API with their own custom query parameters, for example, this can be done.

thewilkybarkid:
:shipit:?

giorgiosironi (approved):
Both visualization of schemas (the top level definition including other sub-schemas) and samples work for me

seanwiseman (approved):

PR #24, opened 2018-05-10
Add API gateway

thewilkybarkid:
Adds a very simple Nginx gateway. A little unnecessary for the skeleton as we're only going to have the one API, but it's a missing layer otherwise.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,6 +21,17 @@ services:
             - ./api-dummy/public:/srv/app/public
         depends_on:
             - api-dummy-fpm
+    api-gateway:
+        image: nginx:1.13
+        ports:
+            - '8085:80'
+        volumes:
+            - ./api-gateway/.docker/nginx-server.conf.template:/etc/nginx/conf.d/default.conf.template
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -21,6 +21,17 @@ services:
             - ./api-dummy/public:/srv/app/public
         depends_on:
             - api-dummy-fpm
+    api-gateway:
+        image: nginx:1.13
+        ports:
+            - '8085:80'
+        volumes:
+            - ./api-gateway/.docker/nginx-server.conf.template:/etc/nginx/conf.d/default.conf.template
+        command: sh -c "envsubst \"`env | awk -F = '{printf \" $$%s\", $$1}'`\" < /etc/nginx/conf.d/default.conf.template > /etc/nginx/conf.d/default.conf && cat /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'"
```
giorgiosironi, line 30: magic but explicit way to specify the configuration; later on the whole configuration file can be mounted if there is an orchestrator

giorgiosironi (approved):

seanwiseman (approved):

PR #26, opened 2018-05-17
Silence missing favicon

giorgiosironi:
The 404 ends up in the nginx container logs, possibly hiding more important messages


thewilkybarkid:
Should do `api-dummy` and `api-gateway` too (not that they'll end up in browsers normally).

giorgiosironi:
Was my thinking, but added them for safety. For example, we have an api-gateway that serves documentation right now.

thewilkybarkid (approved):

Issue #27, opened 2018-05-17
Set up event bus


giorgiosironi:
From https://elifesciences.atlassian.net/browse/ELPP-3620

Basic setup in https://github.com/libero/walking-skeleton/pull/18 as part of ELPP-3622

https://github.com/libero/walking-skeleton/pull/25/files#diff-4e5e90c6228fd48698d074241c2ba760R112 incidentally adds persistence.

giorgiosironi:
Has been done as part of other pull requests?

thewilkybarkid:
Think it's been done.

PR #28, opened 2018-05-17
RFC: Event bus ownership

giorgiosironi:
For #27, the ADR is limited in space, but the reasoning for it follows.

There is an issue of how project sets up (RabbitMQ terminology used here):

- `exchanges` (formerly known as topics in SQS land)
- `queues`

For `queues`, it's usually each consuming project creating its own queue and binding it to one or more existing exchanges (also there's the flexibility to create multiple queues in a project, subscribing to different things, if that's really needed).

For `exchanges`, they can be created by the owner of the data without creating any binding. This leaves the other projects free to bind something to them or not. We avoided that in eLife 2.0 because there is the potential for mutual dependencies, like:

- `metrics` would own the `metrics` exchange, emitting events for journal-cms to consume
    journal-cms would own the podcast-episodes exchange emitting events for metrics to consume (hypothetical, metrics is not consuming events at the moment)

In general, this mutual dependencies between projects didn't happen:

    journal-cms depends on lax but not the other way around
    search depends on lots of things but doesn't emit events
    metrics depends on lots of things but doesn't emit events
    observer depends on lots of things but doesn't emit events
    annotations depends on profiles but not the other way around
    So we could have each project defines its own exchange if we can guarantee an acyclic graph of dependencies between projects (which is a good thing anyway, docker-compose for example agrees).

### DIFF
--- a/adr/0004-event-bus-ownership.md
+++ b/adr/0004-event-bus-ownership.md
@@ -18,9 +18,11 @@ Some services may require a generic binding to many other exchanges in the inter
 
 ## Decision
 
-Create an `exchange` for each data type, as part of a service publishing events.
+Create an `exchange` for each data type or activity, as part of a service publishing events.
 
 Create one `queue` as part of a service consuming events.
+
+Bind each `queue` to as many exchanges as needed, using correlation ids to aggregate events relating to the same data type.
 
 A service may have optional binds, which will only subscribe its `queue` to another service's `exchange` if enabled or dynamically configured as a plugin.
 
--- a/adr/0004-event-bus-ownership.md
+++ b/adr/0004-event-bus-ownership.md
@@ -22,7 +22,7 @@ Create an `exchange` for each data type or activity, as part of a service publis
 
 Create one `queue` as part of a service consuming events.
 
-Bind each `queue` to as many exchanges as needed, using correlation ids to aggregate events relating to the same data type.
+Bind each `queue` to as many exchanges as needed.
 
 If an `exchange` from an upstream service is missing, wait and poll until it's present when bootstrapping a service process that wants to bind to it.
 
--- a/adr/0004-event-bus-ownership.md
+++ b/adr/0004-event-bus-ownership.md
@@ -24,9 +24,11 @@ Create one `queue` as part of a service consuming events.
 
 Bind each `queue` to as many exchanges as needed, using correlation ids to aggregate events relating to the same data type.
 
-The consuming-events-from relationship between services is unidirectional and must be acyclic.
+If an `exchange` from an upstream service is missing, wait and poll until it's present when bootstrapping a service process that wants to bind to it.
 
 A service may have optional binds, which will only subscribe its `queue` to another service's `exchange` if enabled or dynamically configured as a plugin.
+
+The consuming-events-from relationship between services is unidirectional and must be acyclic.
 
 ## Consequences
 
--- a/adr/0004-event-bus-ownership.md
+++ b/adr/0004-event-bus-ownership.md
@@ -1,0 +1,38 @@
+# 4. Own event bus exchanges and queues in services
+
+Date: 2018-05-17
+
+## Status
+
+Proposed
+
+## Context
+
+In a service-based architecture, infrastructure needs to be owned by a service so that it is provisioned and maintained as a first class citizen.
+
+The [event system](0001-events-system.md) based on RabbitMQ requires the setup of `exchanges` and `queues` that `bind` to them.
+
+Practical experience developing eLife lead to no mutual dependencies between services, with e.g. service B strictly being downstream from service A with no events feeding back into service A.
+
+Some services may require a generic binding to many other exchanges in the interest of observability, e.g. a dashboard aggregating events from many different services.
+
+## Decision
+
+Create an `exchange` for each data type, as part of a service publishing events.
+
+Create one `queue` as part of a service consuming events.
+
+Create more than one `queue` in a service if specialized behavior is required, in opposition to a generic type of consumer process.
+
+A service may have optional binds, which will only subscribe its `queue` to another service's `exchange` if enabled or dynamically configured as a plugin.
+
+## Consequences
+
+Service have a [partial order](https://en.wikipedia.org/wiki/Partially_ordered_set) to follow for setup.
+
+The consuming-events-from relationship between services is unidirectional and must be acyclic:
+
+- service B, consuming events from A, cannot publish events for A to consume
+- service C, consuming events from B, also cannot publish events for A to consume
+
+The consuming-events-from relationship will often be part of a stronger dependency, like relying on the APIs exposed by service A.
--- a/adr/0004-event-bus-ownership.md
+++ b/adr/0004-event-bus-ownership.md
@@ -22,8 +22,6 @@ Create an `exchange` for each data type, as part of a service publishing events.
 
 Create one `queue` as part of a service consuming events.
 
-Create more than one `queue` in a service if specialized behavior is required, in opposition to a generic type of consumer process.
-
 A service may have optional binds, which will only subscribe its `queue` to another service's `exchange` if enabled or dynamically configured as a plugin.
 
 ## Consequences
--- a/adr/0004-event-bus-ownership.md
+++ b/adr/0004-event-bus-ownership.md
@@ -24,13 +24,15 @@ Create one `queue` as part of a service consuming events.
 
 Bind each `queue` to as many exchanges as needed, using correlation ids to aggregate events relating to the same data type.
 
+The consuming-events-from relationship between services is unidirectional and must be acyclic.
+
 A service may have optional binds, which will only subscribe its `queue` to another service's `exchange` if enabled or dynamically configured as a plugin.
 
 ## Consequences
 
 Service have a [partial order](https://en.wikipedia.org/wiki/Partially_ordered_set) to follow for setup.
 
-The consuming-events-from relationship between services is unidirectional and must be acyclic:
+The acyclic event consumption implies that:
 
 - service B, consuming events from A, cannot publish events for A to consume
 - service C, consuming events from B, also cannot publish events for A to consume
--- a/events/schema.md
+++ b/events/schema.md
@@ -14,7 +14,7 @@ For clarity, the example data is included with JS syntax (e.g. allowing comments
 {
     "eventId": "448278c4-22d6-11e8-b467-0ed5f89f718b",
     "happenedAt": "2018-03-08T12:00:00+00:00",
-    "aggregate": {
+    "entity": {
         "service": "article-store",
         "name": "article-run",
         "identifier": "6551f09a-bd23-4f12-a039-f8fa78ef776a",
@@ -49,7 +49,7 @@ For clarity, the example data is included with JS syntax (e.g. allowing comments
 {
     "eventId": "448278c4-22d6-11e8-b467-0ed5f89f718b",
     "happenedAt": "2018-03-08T12:00:00+00:00",
-    "aggregate": {
+    "entity": {
         "service": "search",
         "name": "indexing",
         "identifier": "18b21fea-5a87-11e8-9c2d-fa7ae01bbebc",
@@ -89,7 +89,7 @@ For clarity, the example data is included with JS syntax (e.g. allowing comments
 {
     "eventId": "448278c4-22d6-11e8-b467-0ed5f89f718b",
     "happenedAt": "2018-03-08T12:00:00+00:00",
-    "aggregate": {
+    "entity": {
         "service": "downstream-sample",
         "name": "deposit",
         "identifier": "2018-03-08/1", # daily attempts
@@ -133,7 +133,7 @@ Unique for all events. It is a UUID version 1 generated with time and a node ide
 
 Identifies when the event has happened (then it can get published, consumed, etc). It follows the `yyyy-mm-ddTHH:MM:SS+00:00` ISO 8601 format, and is expressed in UTC.
 
-### `aggregate`
+### `entity`
 
 Identifies the unit of consistency that has produced the event: article run in the bot, article version in the articles store, podcast episode in journal-cms, an article's set of metrics in metrics, a single profile in profiles.
 
@@ -155,6 +155,6 @@ Is opaque here, and can contain anything as long as it follows naming convention
 
 Can be used to track events across services, grouping them by the kind of session that originated them or the data type they refer to. Could be an article run during ingestion, or a journal page creating a profile.
 
-Represents an array of dictionaries, each following the format of `aggregate`.
+Represents an array of dictionaries, each following the format of `entity`.
 
 Services should normally add correlation ids from events they have triggered them to the new events being produced.
--- a/events/schema.md
+++ b/events/schema.md
@@ -13,33 +13,10 @@ For clarity, the example data is included with JS syntax (e.g. allowing comments
 ```
 {
     "eventId": "448278c4-22d6-11e8-b467-0ed5f89f718b",
+    "runId": "0509e8c3-4dee-4a3a-800a-3fa9fb43f2e8",
     "happenedAt": "2018-03-08T12:00:00+00:00",
-    "entity": {
-        "service": "article-store",
-        "name": "article-run",
-        "identifier": "6551f09a-bd23-4f12-a039-f8fa78ef776a",
-    },
     "type": "deposit-assets-started",
-    "data": {
-        ... 
-    },
-    "correlationIds": [
-        {
-            "service": "article-store",
-            "name": "article",
-            "identifier": "10627",
-        },
-        {
-            "service": "article-store",
-            "name": "article-version",
-            "identifier": "1",
-        },
-        {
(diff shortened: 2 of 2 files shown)


adr/0004-event-bus-ownership.md
```diff
@@ -0,0 +1,38 @@
...
+
+Some services may require a generic binding to many other exchanges in the interest of observability, e.g. a dashboard aggregating events from many different services.
+
+## Decision
+
+Create an `exchange` for each data type, as part of a service publishing events.
+
+Create one `queue` as part of a service consuming events.
+
+Create more than one `queue` in a service if specialized behavior is required, in opposition to a generic type of consumer process.
```
thewilkybarkid: Might be out of scope for Libero itself?

adr/0004-event-bus-ownership.md
```diff
@@ -0,0 +1,38 @@
...
+
+The [event system](0001-events-system.md) based on RabbitMQ requires the setup of `exchanges` and `queues` that `bind` to them.
+
+Practical experience developing eLife lead to no mutual dependencies between services, with e.g. service B strictly being downstream from service A with no events feeding back into service A.
+
+Some services may require a generic binding to many other exchanges in the interest of observability, e.g. a dashboard aggregating events from many different services.
+
+## Decision
+
+Create an `exchange` for each data type, as part of a service publishing events.
```
thewilkybarkid: Any service can publish to the `exchange`? Thinking Search when starting/finishing indexing an article.

adr/0004-event-bus-ownership.md
```diff
@@ -0,0 +1,38 @@
...
+
+Create more than one `queue` in a service if specialized behavior is required, in opposition to a generic type of consumer process.
+
+A service may have optional binds, which will only subscribe its `queue` to another service's `exchange` if enabled or dynamically configured as a plugin.
+
+## Consequences
+
+Service have a [partial order](https://en.wikipedia.org/wiki/Partially_ordered_set) to follow for setup.
+
+The consuming-events-from relationship between services is unidirectional and must be acyclic:
```
thewilkybarkid: Sounds like a decision.

adr/0004-event-bus-ownership.md
giorgiosironi: So "data type" is too strict, as it should include abstract entitities like a `search indexing run` or similar

giorgiosironi: Option 1, separate exchanges per-project:

- Article Store publishes to `articles`
- Downstream Sample publishes to `downstream-sample`
- Search published to `search`

Different event types require a standard correlation_id to be grouped together. Requires some assembly in the dashboard or anything else listening to multiple sources.

Option 2, rely on the original data type exchange:

- Article Store publishes to `articles` with type `article.published`
- Downstream Sample publishes to `articles` with type `article.crossref.started`
- Search published to `articles` with type `article.search.indexed`

Sightly easier to setup a downstream service, but it will receive and discard its own events. The dependency to the `articles` exchange is already there, so the surface area of a service like Article Store already includes its exchanges (can potentially be wrapped in an HTTP API rather than writing directly to an exchange from a service that doesn't own it). Requires configuration of the data types in the dashboard. Requires active ignoring of all unknown event types as a design decision (wouldn't do RabbitMQ magic with routing keys).

adr/0004-event-bus-ownership.md
giorgiosironi: Just wanted to make clear it's allowed to integrate like that, but can remove.

giorgiosironi:
Going through the various topologies at https://jack-vanlightly.com/blog/2017/12/5/rabbitmq-vs-kafka-part-2-rabbitmq-messaging-patterns-and-topologies I never see loops, so I'd stick by default to applications owning their exchanges. Having queues discard messages seems to be an anti-pattern as exchanges should be able to be route only the messages that are needed.

Case 4 (overkill for us) is interesting in providing an exchange on the consumer side too, to account for custom routing configuration that can change while still being decoupled by the producer's exchange. 

Essentially allows a consumer to use more selective exchanges than the public ones from the producer, or to add rules. Definitely not needed here, but shows that end of the decoupling spectrum.

giorgiosironi:
Summarized this with in-person discussion, ADR would be final upon merge but schema can be revised later if implementations discover problems.

events/schema.md
```diff
@@ -4,30 +4,157 @@ Events published to the bus have a standard schema that allows SDKs and services
...
+
+## Fields
+
+### `eventId`
+
+Unique for all events. It is a UUID version 1 generated with time and a node identifier.
+
+### `happenedAt` 
+
+Identifies when the event has happened (then it can get published, consumed, etc). It follows the `yyyy-mm-ddTHH:MM:SS+00:00` ISO 8601 format, and is expressed in UTC.
```
thewilkybarkid, line 59: Might want to allow milli/microseconds for accuracy?

events/schema.md
```diff
@@ -4,30 +4,157 @@ Events published to the bus have a standard schema that allows SDKs and services
...
+
+### `aggregate`
+
+Identifies the unit of consistency that has produced the event: article run in the bot, article version in the articles store, podcast episode in journal-cms, an article's set of metrics in metrics, a single profile in profiles.
+
+All three information are required to uniquely identify it:
+
+- `service`: `[a-z\-]+`
+- `name`: `[a-z\-]+`
+- `identifier`: `.+` (possibly URL-friendly)
```
thewilkybarkid: For an article version this is a combination of ID and version?

events/schema.md
```diff
@@ -4,30 +4,157 @@ Events published to the bus have a standard schema that allows SDKs and services
...
+
+### `data` (optional)
+
+Is opaque here, and can contain anything as long as it follows naming conventions. Events with the same `type` should usually follow the same schema.
+
+### `correlationIds` (optional)
+
+Can be used to track events across services, grouping them by the kind of session that originated them or the data type they refer to. Could be an article run during ingestion, or a journal page creating a profile.
+
+Represents an array of dictionaries, each following the format of `aggregate`.
```
thewilkybarkid: Ordered?

events/schema.md
```diff
@@ -4,30 +4,157 @@ Events published to the bus have a standard schema that allows SDKs and services
...
+
+Unique for all events. It is a UUID version 1 generated with time and a node identifier.
+
+### `happenedAt` 
+
+Identifies when the event has happened (then it can get published, consumed, etc). It follows the `yyyy-mm-ddTHH:MM:SS+00:00` ISO 8601 format, and is expressed in UTC.
+
+### `aggregate`
+
+Identifies the unit of consistency that has produced the event: article run in the bot, article version in the articles store, podcast episode in journal-cms, an article's set of metrics in metrics, a single profile in profiles.
```
thewilkybarkid: Not sure where this would be used...

events/schema.md
```diff
@@ -4,30 +4,157 @@ Events published to the bus have a standard schema that allows SDKs and services
...
+
+Describes what the event is about so that they can be grouped or recognized. It follows the `[a-z\-\.]+` regular expression, with `.` separating hierarchical levels and `-` separating words in a single level (if there are multiple words).
+
+### `data` (optional)
+
+Is opaque here, and can contain anything as long as it follows naming conventions. Events with the same `type` should usually follow the same schema.
+
+### `correlationIds` (optional)
+
+Can be used to track events across services, grouping them by the kind of session that originated them or the data type they refer to. Could be an article run during ingestion, or a journal page creating a profile.
```
thewilkybarkid: Might be able to solve https://github.com/libero/walking-skeleton/issues/31#issuecomment-390585994 with a consistent `started`/`completed`/`failed` naming in `type` and `aggregate`? So the aggregate is the process, so an 'ingest article' aggregate has a sub-process of 'deposit images', both of these having `started`/`completed` events, and the latter also has the correlation ID of the former.

The dashboard can then construct something like:

- [article.ingest aggregate]
    - article.ingest.start
        - [article.ingest.image aggregate]
            - article.ingest.image.start
            - article.ingest.image.completed
    - article.ingest.completed

events/schema.md
```diff
@@ -4,30 +4,157 @@ Events published to the bus have a standard schema that allows SDKs and services
...
+
+Represents an array of dictionaries, each following the format of `aggregate`.
+
+Services should normally add correlation ids from events they have triggered them to the new events being produced.
```
thewilkybarkid: And on HTTP requests (we'll need a standard for `X-Correlation-Id`).

events/schema.md
thewilkybarkid: Related to https://github.com/libero/walking-skeleton/pull/28/files#r189615065

events/schema.md
giorgiosironi: An `aggregate` in a DDD|Domain Events|Event Sourcing sense is a cluster of objects/rows/documents that can be updated atomically, is colocated and is always internally consistent; for the publishing side an `aggregate` has usually been an `article` or `article version` but in the elife-bot's world it may be an `article run` (which is wider in scope than an `ingest` because it involves the publishing phase too). `run` can be renamed to something else with the same scope, it's just a historical name.

A consistent pattern of `started`|`completed`|`failed` is definitely helpful for handling many heterogeneous events (will add it to the schema).

I see no problem with nested processes, but wouldn't call them aggregates:
```
[article.run entity]
    article.run.start
        [article.run.image entity]
            article.run.image.start
            article.run.image.completed
    article.run.completed
```
casting upon the `aggregate`|`entity` distinction: an `entity` as a object/row/document that is stateful and can have an evolution, and is part of an `aggregate`. Some `entity` in the `aggregate` is the `aggregate root` at its top level. At this point we could say:

1. every event is associated to an `entity` (`article.run.image` or `article.run`)
2. every event is associated to an `aggregate` (`article.run`)

We don't necessarily have to include 2 in the event data if it's not necessary, and/or correlation ids solve the same association problems while remaining more generic. When we have to produce events about a `article.run`, they can be associated to the `article.run` entity.

For simplicity, ids of `entities` can be derived from ids of their `aggregate root` e.g. `fe72663f-8cf3-4e45-9d56-e8a60b0ddeba-image` rather than a whole new UUID.

Slightly renamed example:
```
     "entity": {
         "service": "article-store",
         "name": "article-run",
         "identifier": "6551f09a-bd23-4f12-a039-f8fa78ef776a",
     },
```
```
     "entity": {
         "service": "article-store",
         "name": "article-run-image", # we need some naming conventions here, just using - for continuity until specified
         "identifier": "6551f09a-bd23-4f12-a039-f8fa78ef776a/image", # or .../image/2 if there are many entities like this
     },
```

events/schema.md
giorgiosironi: Can be added, although that use case looks more a specification for distributed tracing rather than for event-based integration between services, and may benefit from a separate look at http://opentracing.io/documentation/pages/spec

events/schema.md
giorgiosironi: Millisecond accuracy on many cloud deployments is not really available, even with ntp:
```
elife@prod--journal:~$ sudo /usr/sbin/ntpdate-debian
23 May 12:28:51 ntpdate[31250]: adjust time server 91.189.89.198 offset -0.478671 sec
```

events/schema.md
giorgiosironi: To be fair there is no clear order, loosely hierarchical or "order of appearance" but there may be cases in which it's ambiguous. Given this is JSON the alternative would be a dictionary of dictionaries, with something being used as key, does that work better?

events/schema.md
giorgiosironi: Basically a logical grouping field with stronger guarantees that `data`, which is schemaless

giorgiosironi: (the equivalent to `id` in https://github.com/elifesciences/bus)

events/schema.md
thewilkybarkid: Think it's an unordered set, so array of objects is fine (just documentation that the order isn't important). Unless there's a natural key, which there isn't (being value objects).

events/schema.md
```diff
@@ -4,30 +4,159 @@ Events published to the bus have a standard schema that allows SDKs and services
...
+
+Is opaque here, and can contain anything as long as it follows naming conventions. Events with the same `type` should usually follow the same schema.
+
+### `correlationIds` (optional)
+
+Can be used to track events across services, grouping them by the kind of session that originated them or the data type they refer to. Could be an article run during ingestion, or a journal page creating a profile.
+
+Represents an array of dictionaries, each following the format of `entity`.
+
+Services should normally add correlation ids from events they have triggered them to the new events being produced.
```
thewilkybarkid: Looking at https://github.com/libero/walking-skeleton/pull/28#discussion_r190211098, _might_ be a replacement?

giorgiosironi: OpenTracing is specific to tracing a tree of remote procedure calls on many different nodes, so it's more technical in nature than business oriented (for example it would include load balancers). The correlation ids used in domain events happen to have a similar name, but should be related to entities that have some meaning for a dashboard for example.

It definitely tells us not to go into more detail than a transaction per service (because finer-grained details is for developers debugging, not dashboard users or integrating services):

- `article` and `article version` are general business-relevant information
- `article run` is the main transaction for the `article store`
- `search indexing` is the main transaction for a downstream service
- `deposit` is the main transaction for another downstream service

Going to capture this to avoid proliferation of correlation ids

events/schema.md
```diff
@@ -4,30 +4,168 @@ Events published to the bus have a standard schema that allows SDKs and services
 
 The schema is based on JSON (and can be modelled with JSON Schema if necessary).
```
thewilkybarkid, line 5: (Unrelated, but since there's quite a bit of rethinking happening here) thought: should we follow the API's RELAX NG setup for events? JSON is definitely simpler, and as long as inline content isn't passed around (thinking article titles might appear in `data`) it's not a problem. But, it would provide the same extensibility model and (strong) validation.

thewilkybarkid:
Trying to think of how the dashboard could show what's going on in a sensible fashion. We should really talk to business people about what's most useful, but the scope of the dashboard should probably be increased to be more technical (for certain users, ie us).

eLife's current dashboard shows things in a linear fashion (even though it can technically do things in parallel?) Since we're talking about a DAG here, it should be possible to visualise it as a graph or timeline (like in [the OpenTracing spec](https://github.com/opentracing/specification/blob/master/specification.md#the-opentracing-data-model)). It's a little different to OpenTracing, though, in that each span (just to use their terminology, transaction might be better) is made up of at least 2 events (`started` and `completed`/`failed`, and possibly progress events in the middle, eg slow FTP of a large ZIP shouldn't timeout on the dashboard). Spans can have sub-spans. Spans can also follow other spans.

Currently it seems hard to group them like that. Information should exist in `correlationIds`, but that's currently optional and has to be interpreted. Should we take more of an opinionated approach to the event bus and make it explicit that every event is part of a span? (And provide OpenTracing-like `ChildOf` and `FollowsFrom` in the details about the span?)

Means events can't be standalone, but don't think that's needed? Most services can just listen for `*.completed` events.

events/schema.md
giorgiosironi: IIRC the original bus events in Continuum didn't contain any `data` apart from identifiers not to duplicate what is provided by the API; now we are introducing `data` and possibly a schema for it, which veers towards Domain Events/CQRS/Event Sourcing (but not necessarily persisting events in the long term, which simplifies schema changes). My main concern is how easy an event should be to produce/consume for a random service.

giorgiosironi:
> And provide OpenTracing-like ChildOf and FollowsFrom in the details about the span?

`FollowsFrom` is the one used for asynchronous, bus-based interactions where the outcome of the producer is not related to the outcome of the consumer

> Trying to think of how the dashboard could show what's going on in a sensible fashion. We should really talk to business people about what's most useful, but the scope of the dashboard should probably be increased to be more technical (for certain users, ie us).

I assumed technical users could work with opaque data (e.g. a dictionary dump for each event) as it's analogue to structured logging; without knowing more about the UX requirements it's difficult to say how semantically rich should the events be.

giorgiosironi:
- add some correlation-related field to events that can be used to refer to a previous event that spawned it
- add some standard for transaction-based events: types; it is unclear whether such a standard could exist for entity-based events, maybe it will emerge

giorgiosironi:
- events containing data (such as properties article id, title, authors) may be emitted by many different upstream services (article store but also outside of Libero) so they should contain data that is related to the transaction
- we need some kind of standard for these data, from a JSON dictionary to something more complex depending on the requirement

giorgiosironi:
Updated event schema with the simplification.

adr/0004-event-bus-ownership.md
```diff
@@ -0,0 +1,42 @@
...
+
+Some services may require a generic binding to many other exchanges in the interest of observability, e.g. a dashboard aggregating events from many different services.
+
+## Decision
+
+Create an `exchange` for each data type or activity, as part of a service publishing events.
+
+Create one `queue` as part of a service consuming events.
+
+Bind each `queue` to as many exchanges as needed, using correlation ids to aggregate events relating to the same data type.
```
thewilkybarkid: Needs updating.

thewilkybarkid (approved):

Issue #31, opened 2018-05-21
Article store events publisher


seanwiseman:
Add functionality to publish article events via the walking skeleton's [internal event bus](https://github.com/libero/walking-skeleton/blob/master/adr/0001-events-system.md).

Should include the publishing of:
- [x] Article version created `started/succeeded/failed` events
- [x] Article version update `started/succeeded/failed` events
- [x] Article deleted `succeeded/failed` events

thewilkybarkid:
For now we could have:

- `article.version.[delete|post|put].started`
- `article.version.[delete|post|put].completed`
- `article.version.[delete|post|put].failed`

Ties to the API (in name), but this _is_ the processing of the request...

Not sure about the last two though: ok for this purpose, but more generally we might want to consider modelling processes and sub-processes (ie start, step (...), result) in the event bus, rather than just standalone events?

seanwiseman:
OK sounds good. Yes modelling them would also give a nicer context and would perhaps make it simpler to manage any side effects, such as a failure on step `n`.

PR #25, opened 2018-05-16
ELPP-3621 Downstream service sample: emit one event in response to article

giorgiosironi:
For #29 

See README for how much is implemented in this PR and what will follow.

Sample log:
```
downstream-sample_1                | Receiving: {'eventId': '7d5de096-6338-11e8-83a9-0242ac1a000c', 'happenedAt': '2018-05-29T12:04:50+00:00', 'aggregate': {'service': 'article-store', 'name': 'article-version', 'identifier': '0065'}, 'type': 'article.version.post.started', 'data': {'article_version': 1}}
downstream-sample_1                | Publishing: {'runId': None, 'type': 'downstream.crossref.started', 'message': 'We are happy to receive this paper and have put it into a queue', 'eventId': '72d65f3a-6339-11e8-b721-0242ac1a000a', 'happenedAt': '2018-05-29T12:11:42+00:00'}
downstream-sample_1                | Receiving: {'eventId': '7d6ef94e-6338-11e8-83a9-0242ac1a000c', 'happenedAt': '2018-05-29T12:04:50+00:00', 'aggregate': {'service': 'article-store', 'name': 'article-version', 'identifier': '0065'}, 'type': 'article.version.post.completed', 'data': {'article_version': 1}}
downstream-sample_1                | Publishing: {'runId': None, 'type': 'downstream.crossref.started', 'message': 'We are happy to receive this paper and have put it into a queue', 'eventId': '72d8925a-6339-11e8-b721-0242ac1a000a', 'happenedAt': '2018-05-29T12:11:42+00:00'}
```

PR #32, opened 2018-05-22
Article store events

seanwiseman:
Implements requirements from [#31](https://github.com/libero/walking-skeleton/issues/31).

### DIFF
--- a/article-store/Pipfile
+++ b/article-store/Pipfile
@@ -10,12 +10,14 @@ name = "pypi"
 django = "~=2.0.5"
 django-debug-toolbar = "~=1.9.1"
 djangorestframework = "~=3.8.2"
-psycopg2-binary = "*"
+"psycopg2-binary" = "*"
 django-filter = "~=1.1.0"
 requests = "~=2.18.4"
 djangorestframework-xml = "~=1.3.0"
 lxml = "~=4.2.1"
 dj-database-url = "~=0.5.0"
+pika = "~=0.11.2"
+
 
 [dev-packages]
 
--- a/article-store/articles/tests/test_api/test_article_api.py
+++ b/article-store/articles/tests/test_api/test_article_api.py
@@ -1,4 +1,5 @@
 import pytest
+from unittest.mock import patch
 
 from django.test.client import Client
 
@@ -12,6 +13,13 @@ ARTICLES_URL = '/articles'  # TODO refactor using this const!
 
 
 class TestCreateArticleVersion:
+
+    def setup(self):
+        self.mock_publisher = patch('articles.api.message_publisher')
+        self.mock_publisher.start()
+
+    def tearDown(self):
+        self.mock_publisher.stop()
 
     @pytest.mark.django_db
     def test_can_create_article_version(self, admin_client: Client, article: Article):
--- a/article-store/articles/tests/test_api/test_article_api.py
+++ b/article-store/articles/tests/test_api/test_article_api.py
@@ -15,7 +15,7 @@ class TestCreateArticleVersion:
 
     @pytest.mark.django_db
     def test_can_create_article_version(self, admin_client: Client, article: Article):
-        response = admin_client.post(f'{ARTICLES_URL}/{article.id}',
+        response = admin_client.post(f'{ARTICLES_URL}/{article.id}/versions',
                                      data='', content_type='application/xml')
         assert response.status_code == 201
         assert ArticleVersion.objects.count() == 1
@@ -23,7 +23,7 @@ class TestCreateArticleVersion:
     @pytest.mark.django_db
     def test_will_not_accept_invalid_article_id_format(self, admin_client: Client):
         invalid_id = '[foo](bar)**'
-        response = admin_client.post(f'{ARTICLES_URL}/{invalid_id}',
+        response = admin_client.post(f'{ARTICLES_URL}/{invalid_id}/versions',
                                      data='', content_type='application/xml')
         assert response.status_code == 406
 
@@ -31,7 +31,7 @@ class TestCreateArticleVersion:
     def test_can_create_article_version_with_content(self, admin_client: Client,
                                                      article: Article,
                                                      content_xml_data: str):
-        response = admin_client.post(f'{ARTICLES_URL}/{article.id}',
+        response = admin_client.post(f'{ARTICLES_URL}/{article.id}/versions',
                                      data=content_xml_data,
                                      content_type='application/xml')
         assert response.status_code == 201
@@ -47,7 +47,7 @@ class TestUpdateArticleVersion:
                                                      article_version_1: ArticleVersion,
                                                      content_en_front: Content,
                                                      content_xml_data: str):
-        response = admin_client.put(f'{ARTICLES_URL}/{article.id}/{article_version_1.version}',
+        response = admin_client.put(f'{ARTICLES_URL}/{article.id}/versions/{article_version_1.version}',
                                     data=content_xml_data,
                                     content_type='application/xml')
         assert response.status_code == 201
@@ -62,7 +62,7 @@ class TestDeleteArticles:
                                         admin_client: Client, article: Article,
                                         article_version_1: ArticleVersion,
                                         content_en_front: Content):
-        url = f'/articles/{article.id}/{article_version_1.version}'
+        url = f'/articles/{article.id}/versions/{article_version_1.version}'
         response = admin_client.delete(url)
         assert response.status_code == 202
         assert Article.objects.count() == 1
@@ -78,7 +78,7 @@ class TestDeleteArticles:
                                                      content_en_front: Content):
         assert ArticleVersion.objects.count() == 3
 
-        url = f'/articles/{article.id}/{article_version_2.version}'
+        url = f'/articles/{article.id}/versions/{article_version_2.version}'
         response = admin_client.delete(url)
         assert response.status_code == 202
         assert Article.objects.count() == 1
@@ -127,7 +127,7 @@ class TestArticleErrors:
 
     @pytest.mark.django_db
     def test_can_handle_article_not_existing(self, admin_client: Client):
-        url = '/articles/cc1250b5-0855-4fc2-906f-a6a77e4c90f5/latest'
+        url = '/articles/cc1250b5-0855-4fc2-906f-a6a77e4c90f5/versions/latest'
         response = admin_client.get(url)
         assert response.status_code == 406
         assert 'Article matching query does not exist' in response.content.decode('utf-8')
@@ -136,7 +136,7 @@ class TestArticleErrors:
     def test_can_handle_no_article_version_present(self,
                                                    admin_client: Client,
                                                    article: Article):
-        url = f'/articles/{article.id}/latest'
+        url = f'/articles/{article.id}/versions/latest'
         response = admin_client.get(url)
         assert response.status_code == 406
         assert 'Content does not exist' in response.content.decode('utf-8')
@@ -146,7 +146,7 @@ class TestArticleErrors:
                                                            admin_client: Client,
                                                            article: Article,
                                                            article_version_1: ArticleVersion):
-        url = f'/articles/{article.id}/latest'
+        url = f'/articles/{article.id}/versions/latest'
         response = admin_client.get(url)
         assert response.status_code == 406
         assert 'Content does not exist' in response.content.decode('utf-8')
@@ -161,7 +161,7 @@ class TestArticleContent:
                                      article_version_1: ArticleVersion,
                                      content_en_front: Content,
                                      article_0065_en_front_xml: str):
-        url = f'/articles/{article.id}/{article_version_1.version}/{content_en_front.name}'
+        url = f'/articles/{article.id}/versions/{article_version_1.version}/{content_en_front.name}'
         response = admin_client.get(url)
         assert response.status_code == 200
         xml_payload = response.content.decode('utf-8')
@@ -175,7 +175,7 @@ class TestArticleContent:
                                                               content_en_front: Content,
                                                               article_0065_en_front_xml: str):
 
-        url = f'/articles/{article.id}/latest/{content_en_front.name}'
+        url = f'/articles/{article.id}/versions/latest/{content_en_front.name}'
         response = admin_client.get(url)
         assert response.status_code == 200
         assert response.content.decode('utf-8') == article_0065_en_front_xml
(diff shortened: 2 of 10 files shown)


seanwiseman:
Currently sends messages for the following events:

`article.delete.[started|completed|failed]`
`article.version.[delete|post|put].[started|completed|failed]`

You should see them populate the dashboard.

dashboard/config/packages/old_sound_rabbit_mq.yaml
```diff
@@ -8,6 +8,6 @@ old_sound_rabbit_mq:
             enable_logger: true
             exchange_options:
                 name: 'dashboard'
-                type: direct
+                type: fanout
             queue_options:
-                name: 'dashboard'
+                name: 'articles'
```
thewilkybarkid: Should be the exchange's name changed?

seanwiseman: As we are not producing any messages (from what I can see) from the dashboard at present, I am not sure of the exchange's purpose here 🤔 .  Maybe was just for testing.

You only need the queue name (`articles` in this case) to receive the messages. The `exchange_options` entry can be removed if we are not producing messages from the dashboard.

thewilkybarkid: Should be the exchange that the queue is bound to (this is under the `consumers` config rather than `producers`).

article-store/events/conf.py
```diff
@@ -0,0 +1,21 @@
...
+import pika
+
+
+RABBITMQ_HOST = os.environ.get('RABBITMQ_HOST', 'rabbitmq')  # event-bus
+RABBITMQ_PORT = os.environ.get('RABBITMQ_PORT', 5672)
+RABBITMQ_PASSWORD = os.environ.get('RABBITMQ_PASSWORD')
+RABBITMQ_USER = os.environ.get('RABBITMQ_USER')
+
+BROKER_CREDENTIALS = pika.PlainCredentials(RABBITMQ_USER, RABBITMQ_PASSWORD)
+BROKER_PARAMS = pika.ConnectionParameters(host=RABBITMQ_HOST, credentials=BROKER_CREDENTIALS)
```
thewilkybarkid: Same as https://github.com/libero/walking-skeleton/pull/25#discussion_r189183325

article-store/events/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    try:
+        connection = pika.BlockingConnection(parameters=BROKER_PARAMS)
+        yield connection.channel()
+        connection.close()
+    except ConnectionClosed as err:
+        print(err)
+        raise
+
+
+def setup_exchanges_and_queues(func) -> Callable[..., None]:
```
thewilkybarkid: We don't need a queue here? Though guess this is just copied from earlier work?

dashboard/config/packages/old_sound_rabbit_mq.yaml
thewilkybarkid: Can see this is actually working... taking a proper look.

article-store/events/utils.py
seanwiseman: Yes, it was brought in from previous work, but also makes sure the queue(s) you are binding the exchange(s) to exists before the actual binding. This may be subject to change if we can guarantee a queues existence but for now, I think we have to do it this way.

thewilkybarkid: But in this case we don't need the `articles` queue?

article-store/events/conf.py
seanwiseman: Done 👍

article-store/events/utils.py
seanwiseman: No, it only actually needs an exchange to publish.

seanwiseman:
Are we going to want the `article.version.published` event available in this PR?

article-store/Pipfile
```diff
@@ -10,12 +10,14 @@ name = "pypi"
 django = "~=2.0.5"
 django-debug-toolbar = "~=1.9.1"
 djangorestframework = "~=3.8.2"
-psycopg2-binary = "*"
+"psycopg2-binary" = "*"
```
giorgiosironi, line 13: pipenv strikes back - not a problem

article-store/articles/tests/test_api/test_article_api.py
```diff
@@ -227,7 +235,7 @@ def test_can_get_article_content_without_giving_part_name(self,
                                                               content_es_front: Content,
                                                               article_0065_es_front_xml: str):
 
-        url = f'/articles/{article.id}/latest'
+        url = f'/articles/{article.id}/versions/latest'
```
giorgiosironi, line 238: changes like these should usually be a separate PR?

giorgiosironi: not that the history of the walking skeleton itself should be that clean in the long run, projects are going to be extracted in other repositories

article-store/events/conf.py
```diff
@@ -0,0 +1,17 @@
+import os
+
+import pika
+
+
+RABBITMQ_URL = os.environ.get('RABBITMQ_URL')
+
+BROKER_PARAMS = pika.connection.URLParameters(RABBITMQ_URL)
+
+ARTICLES_QUEUE = os.environ.get('ARTICLES_QUEUE', 'articles')
```
giorgiosironi: I agree with other comments that a queue should not be necessary here. If we want to set up one for testing, it should be clearly named as `articles-dummy` or similar

article-store/events/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    :param article_id: str
+    :param article_version: int
+    :return:
+    """
+    with get_channel() as channel:
+        message = create_message(msg_type=msg_type, identifier=article_id,
+                                 data={'article_version': article_version})
+
+        channel.basic_publish(exchange=ARTICLES_EXCHANGE,
+                              routing_key="",
```
giorgiosironi, line 92: didn't understand this can be empty, good to know

article-store/articles/tests/test_api/test_article_api.py
seanwiseman: Yes I agree.

article-store/events/utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+def create_message(msg_type: str, identifier: str, data: Dict) -> Dict:
+    """Create a message `dict` based on a standard schema.
+
+    :return: Dict
+    """
+    return {
+        "eventId": str(uuid.uuid1()),
+        "happenedAt": strftime("%Y-%m-%dT%H:%M:%S+00:00", gmtime()),
+        "aggregate": {
```
giorgiosironi, line 35: may be changed to `entity` in the schema (not yet)

article-store/events/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    :return:
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        with get_channel() as channel:
+            for queue_name in DEFAULT_QUEUES:
+                channel.queue_declare(queue=queue_name, durable=True)
+
+                for exchange in DEFAULT_QUEUES[queue_name]:
+                    channel.exchange_declare(exchange=exchange, exchange_type='fanout', durable=True)
```
giorgiosironi: exchanges (only `ARTICLES_EXCHANGE`) should be declared even if there are no queues to be bound to them; this code can be pulled up to the start of the `with get_channel()` block

article-store/events/utils.py
```diff
@@ -0,0 +1,122 @@
...
+def message_publisher(msg_type: str, article_id: str, article_version: int = None) -> ContextManager[None]:
+    """Wrapper to send started/completed/failed messages for a given article `msg_type`.
+
+    :param msg_type: str
+    :param article_id: str
+    :param article_version: int
+    :return:
+    """
+
+    send_message = partial(send_article_message, article_id=article_id, article_version=article_version)
```
giorgiosironi, line 107: this is neat

giorgiosironi: (this context manager in general)

article-store/events/utils.py
```diff
@@ -0,0 +1,122 @@
...
+    """Create a message `dict` based on a standard schema.
+
+    :return: Dict
+    """
+    return {
+        "eventId": str(uuid.uuid1()),
+        "happenedAt": strftime("%Y-%m-%dT%H:%M:%S+00:00", gmtime()),
+        "aggregate": {
+            "service": "article-store",
+            "name": "article-version",
```
giorgiosironi, line 37: may be `article` if this is an `article.delete`-related event?

seanwiseman: mmm, the `article.delete` or `article.version.delete` is currently defined in the `type`, does `aggregate.name` have to change with it as well? or could `aggregate.name` just always be `article`?

dashboard/config/packages/old_sound_rabbit_mq.yaml
thewilkybarkid: I had worked out what was happening, must not have submitted my follow-up. 🤔

This results in an exchange called 'dashboard' and a bound queue called 'articles', the article store _was_ creating an exchange called 'articles' and a bound queue called 'articles'. This works as the queues matched, the 'dashboad' exchange was just superfluous.

Now the article store doesn't create a queue, this probably no longer works. `exchange_options.name` needs to be  'articles' and `queue_options.name` needs to be 'dashboard'. That binds the queue for this app to the exchange provided by the article store.

article-store/events/utils.py
thewilkybarkid: Agreed!

article-store/events/utils.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+
+    send_message = partial(send_article_message, article_id=article_id, article_version=article_version)
+
+    try:
+        send_message(f'{msg_type}.started')
+        yield
+        send_message(f'{msg_type}.completed')
+    except Exception:
+        send_message(f'{msg_type}.failed')
```
thewilkybarkid, line 114: Shouldn't this `raise`?

article-store/events/conf.py
```diff
@@ -3,15 +3,12 @@
 import pika
 
 
-RABBITMQ_URL = os.environ.get('RABBITMQ_URL')
+RABBITMQ_URL = os.environ.get('RABBITMQ_URL', '')
 
-BROKER_PARAMS = pika.connection.URLParameters(RABBITMQ_URL, '')
+BROKER_PARAMS = pika.connection.URLParameters(RABBITMQ_URL)
 
 ARTICLES_DUMMY_QUEUE = os.environ.get('ARTICLES_DUMMY_QUEUE', 'articles-dummy')
```
thewilkybarkid: Can be removed?

article-store/events/utils.py
seanwiseman: Good spot.

dashboard/config/packages/old_sound_rabbit_mq.yaml
seanwiseman: Ok, I have made the change and retested and it works fine.

article-store/Pipfile
thewilkybarkid: https://github.com/pypa/pipenv/issues/1933

article-store/events/utils.py
giorgiosironi: I'd say comment out the `aggregate` field for now and wait for https://github.com/libero/walking-skeleton/pull/28/files to add it later when stable

seanwiseman: Ok, the article id is in `aggregate.identifier` currently. Should it be moved elsewhere for the time being?

giorgiosironi: No need to do big changes, just adding a comment here about this field being experimental and about to evolve/being replaced would be fine.

dashboard/config/packages/old_sound_rabbit_mq.yaml
```diff
@@ -7,7 +7,7 @@ old_sound_rabbit_mq:
             callback: Libero\Dashboard\Consumer\EventsConsumer
             enable_logger: true
             exchange_options:
-                name: 'dashboard'
-                type: direct
+                name: 'articles'
+                type: fanout
```
giorgiosironi, line 11: not sure why the dashboard should know about the type of the exchange, is it required to subscribe to it? The risk is a race condition where the dashboard preempts the article store in creating the exchange, which shouldn't belong to that service.

thewilkybarkid: Untested, but looks like you can set `declare: false` and it won't create it.

seanwiseman: Trying to test this now. Which level is `declare` configured at in the file?

All combinations I have tried so far have still lead to `You must provide an exchange type  
`.

thewilkybarkid: Same level (https://github.com/php-amqplib/RabbitMqBundle/blob/cf67adaa4e306d8e9cb6855a72d79263b425ded8/DependencyInjection/Configuration.php#L371-L389), so should be:

```yaml
exchange_options:
    name: 'articles'
    type: fanout
    declare: false
```

seanwiseman: Ah ok, thank you. I thought the `declare` meant we did not have to specify a value for the `type`.

OK, this works. It does not create the exchange when `declare: false` is present.

thewilkybarkid: I was surprised too, feels redundant.

seanwiseman: This however causes the following condition: (after a clean startup with no previous volume data)

- `dasboard` queue present
- `articles` exchange not present until the store creates it
- `dashboard` therefore cannot bind to it: (error on startup)
```
In AMQPChannel.php line 188:

NOT_FOUND - no exchange 'articles' in vhost '/'  
```
- There does not seem to be another attempt to bind (though are we expecting it to?)
- Even after a manual binding (via the rabbitMQ UI) of `dashboard` --> `articles` and new events being sent, the dashboard (app) still does not consume the events. I can however see the events pending on the queue (via the rabbitMQ UI) so they are being delivered to the `dashboard` queue.

Will probably benefit from an in person discussion.

giorgiosironi: Yes, we've seen this before https://github.com/libero/walking-skeleton/pull/25/files#diff-bca2c15ea701ffde5b33224920caa539R39

seanwiseman: As discussed, I will not implement the `declare: false`, it will be handled in another pr along with a method of checking the exchange is declared before trying to bind.

seanwiseman:
:shipit: ?

giorgiosironi (approved):
Event data needs to be removed/refined, but let's do it separately not to hold this

thewilkybarkid (approved):
🎉

PR #34, opened 2018-05-23
Add recommendation for --renew-anon-volumes

thewilkybarkid:
This seems to be the 'solution' for the unmounting-inside-volumes-being-persistent problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ It contains:
 
 ## Applications
 
-These can be run using `docker-compose up`.
+These can be run using `docker-compose up -V`.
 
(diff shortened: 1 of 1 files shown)


seanwiseman (approved):

giorgiosironi (approved):
Workaround should be included at the moment, I suspect there will be more problems in this area

thewilkybarkid:
Yep, still looking for better solutions that are both container and non-container compatible.

giorgiosironi:
Related non-documented behavior, for reference: https://stackoverflow.com/a/47873699/91590

PR #35, opened 2018-05-23
Add Airflow services to project level docker-compose.yml

seanwiseman:


docker-compose.yml
```diff
@@ -1,6 +1,73 @@
 version: '3'
 
 services:
+    airflow-flower:
```
thewilkybarkid: Should probably prefix these all with `article-store-`

docker-compose.yml
```diff
@@ -1,6 +1,73 @@
...
+            - airflow-webserver
+        volumes:
+            - './workflow/airflow_libero/dags:/usr/local/airflow/dags'
+            - './workflow/airflow_libero/plugins:/usr/local/airflow/plugins'
+        environment:
+            - LOAD_EX=n
+            - FERNET_KEY=46BKJoQYlPPOexq0OhDZnIlNepKFf87WFwLbfzqDDho=
+            - EXECUTOR=Celery
+        command: scheduler
+    airflow-webserver:
```
thewilkybarkid: This is just the UI? Could not include it?

docker-compose.yml
```diff
@@ -1,6 +1,73 @@
...
+            timeout: 30s
+            retries: 3
+    airflow-worker:
+        build:
+          context: ./workflow/airflow_libero
+        depends_on:
+            - airflow-scheduler
+        volumes:
+            - './workflow/airflow_libero/dags:/usr/local/airflow/dags'
+            - './workflow/airflow_libero/plugins:/usr/local/airflow/plugins'
```
thewilkybarkid: Interesting question where these should live. Maybe the article store should provide some, and it be possible to add more of your own?

docker-compose.yml
seanwiseman: Certainly, do not 'need' it, but it is very useful whilst testing/experimenting. I think the benefit outweighs the cost of the container for the skeleton.

thewilkybarkid: Could add it to a `docker-compose.override.yml`. Though I'm probably just overthinking.

docker-compose.yml
seanwiseman: Could be anywhere, I think I will make the change to move them somewhere in the next `article-store` PR.

Past the walking-skeleton, I think a separate package with the custom reusable tasks and DAGs is a good option. That way it can just be imported as a dependency.

docker-compose.yml
seanwiseman: `article-store-airflow-*` ?

docker-compose.yml
```diff
@@ -64,6 +64,73 @@ services:
...
+          context: ./workflow/airflow_libero
+        depends_on:
+            - article-store-airflow-redis
+        environment:
+            - EXECUTOR=Celery
+        ports:
+            - "5555:5555"
+        command: flower
+    article-store-airflow-postgres:
+        image: postgres:9.6
```
thewilkybarkid: Can we make `postgres:9.6.8-alpine` to match the others?

docker-compose.yml
```diff
@@ -64,6 +64,73 @@ services:
...
+        environment:
+            - POSTGRES_USER=airflow
+            - POSTGRES_PASSWORD=airflow
+            - POSTGRES_DB=airflow
```
thewilkybarkid: Does Airflow have these defaults?

seanwiseman: Yes, set [here](https://github.com/libero/walking-skeleton/blob/3c38fbc62638825dd9ea8a60386fe3564824651b/workflow/airflow_libero/script/entrypoint.sh#L11), so can be changed if need be really. Was just following some example deployments.

docker-compose.yml
```diff
@@ -64,6 +64,73 @@ services:
...
+            - 'article-store-airflow-postgres:/var/lib/postgresql/data'
+    article-store-airflow-redis:
+        image: 'redis:3.2.7'
+    article-store-airflow-scheduler:
+        build:
+          context: ./workflow/airflow_libero
+        depends_on:
+            - article-store-airflow-webserver
+        volumes:
+            - './workflow/airflow_libero/dags:/usr/local/airflow/dags'
```
thewilkybarkid, line 94: There's a few things in play here:

1. Containerised Airflow
2. Containerised Libero Airflow
3. Libero DAGs
4. Custom DAGs

I'm interested to see how this could look separated. (1 might be ignored depending on how 2 works.)

seanwiseman: You can probably ignore 1, really Its just containerised Libero Airflow and the need for x2 volumes (plugins and dags).

In the article-store, airflow integration PR I will be stripping out bits from the `libero_airflow` package and making it only contain the config bits so it should tidy it up.

workflow/airflow_libero/script/entrypoint.sh
```diff
@@ -2,11 +2,11 @@
 
 TRY_LOOP="20"
 
-: "${REDIS_HOST:="redis"}"
+: "${REDIS_HOST:="article-store-airflow-redis"}"
```
giorgiosironi: these defaults are not used very much, because running an image alone will try to connect to a `redis` or `postgres` container that has not been started instead. Running the image within docker-compose will make it connect to a started `redis` or `postgres`, but will also pass `REDIS_HOST`|`POSTGRES_HOST` so it will overwrite this. Is there a more useful default value to use here? Empty? Localhost?

docker-compose.yml
```diff
@@ -64,6 +64,73 @@ services:
...
+            - './workflow/airflow_libero/plugins:/usr/local/airflow/plugins'
+        environment:
+            - LOAD_EX=n
+            - FERNET_KEY=46BKJoQYlPPOexq0OhDZnIlNepKFf87WFwLbfzqDDho=
```
giorgiosironi, line 98: sounds like a variable for `.env` as it is duplicated across the configuration and can have a user-provided value

seanwiseman: I agree. Perhaps leave this for a follow up when we create an .env for these kinds of variables and also the range of ports being used across containers (from our discussion yesterday).

workflow/airflow_libero/script/entrypoint.sh
seanwiseman: You are right, I will switch them to localhost.

giorgiosironi (approved):

docker-compose.yml
```diff
@@ -64,6 +64,73 @@ services:
                 '
         depends_on:
             - article-store-postgres
+    article-store-airflow-flower:
+        build:
+          context: ./workflow/airflow_libero
```
thewilkybarkid, line 69: Could move this up to the top level (not an experiment anymore).

thewilkybarkid (approved):

PR #36, opened 2018-05-25
Don't create the articles exchange in the dashboard

thewilkybarkid:
Currently the `articles` exchange isn't created until a message is sent, so the queue will miss the first message.

### DIFF
--- a/dashboard/config/packages/old_sound_rabbit_mq.yaml
+++ b/dashboard/config/packages/old_sound_rabbit_mq.yaml
@@ -9,5 +9,6 @@ old_sound_rabbit_mq:
             exchange_options:
                 name: 'articles'
                 type: fanout
+                declare: false
             queue_options:
                 name: 'dashboard'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -145,7 +145,11 @@ services:
             - APP_SECRET=f6a87c2e031ecadc0d7eb442729435c7
             - DATABASE_URL=pgsql://postgres:@dashboard-postgres:5432/postgres
             - RABBITMQ_URL=amqp://guest:guest@event-bus:5672
-        command: bash -c 'wait-for-port event-bus 5672 && bin/console rabbitmq:consumer events'
+        command: >
+                /bin/bash -c '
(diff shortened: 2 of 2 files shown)


giorgiosironi:
> Currently the articles exchange isn't created until a message is sent

That's a bug in the article store?

dashboard/config/packages/old_sound_rabbit_mq.yaml
```diff
@@ -9,5 +9,6 @@ old_sound_rabbit_mq:
             exchange_options:
                 name: 'articles'
                 type: fanout
+                declare: false
```
giorgiosironi, line 12: correct not to declare it from here

docker-compose.yml
```diff
@@ -145,7 +145,11 @@ services:
             - APP_SECRET=f6a87c2e031ecadc0d7eb442729435c7
             - DATABASE_URL=pgsql://postgres:@dashboard-postgres:5432/postgres
             - RABBITMQ_URL=amqp://guest:guest@event-bus:5672
-        command: bash -c 'wait-for-port event-bus 5672 && bin/console rabbitmq:consumer events'
+        command: >
```
giorgiosironi, line 168: it doesn't scale to deal with startup in docker-compose configuration, but it's still easier to do it synchronously than to run another "migration"/"setup" container

seanwiseman:
> Currently the articles exchange isn't created until a message is sent, so the queue will miss the first message.

'Might' miss the first message, as the exchange is created and then a message is sent. Between the creation and sending a bind could happen, though granted its unlikely.

> That's a bug in the article store?

The creation/check of the exchange before sending is really just a precaution, perhaps we need to formalize the creation of the exchange to make it more predictable? we can't really have messages being missed.

giorgiosironi:
> really just a precaution, perhaps we need to formalize the creation of the exchange to make it more predictable?

Should happen on bootstrap or before the application is started imho, like for the relational database schema.

seanwiseman:
Agreed. I will add that in.

giorgiosironi (approved):
The description is out of the date, but the intended behavior is correct

PR #37, opened 2018-05-25
Add exchange setup

seanwiseman:
Relates to [#36](https://github.com/libero/walking-skeleton/pull/36).

...and [this comment](https://github.com/libero/walking-skeleton/pull/36#issuecomment-392079659).

Allows for the creation of the `DEFAULT_EXCHANGES` via a `manage.py` command via the article-store. This then allows you to include the setup as a `docker-compose` service to make sure the `articles` exchange is defined when the services are brought up.

### DIFF
--- a/article-store/events/utils.py
+++ b/article-store/events/utils.py
@@ -42,6 +42,16 @@ def create_message(msg_type: str, identifier: str, data: Dict) -> Dict:
     }
 
 
+def declare_exchanges() -> None:
+    """Declare all default exchanges.
+
+    :return:
+    """
+    with get_channel() as channel:
+        for exchange in DEFAULT_EXCHANGES:
+            channel.exchange_declare(exchange=exchange, exchange_type='fanout', durable=True)
+
+
 @contextmanager
 def get_channel() -> ContextManager[BlockingChannel]:
     """Handles the creation and clean up of a connection,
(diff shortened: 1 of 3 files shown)


article-store/events/utils.py
```diff
@@ -66,10 +76,7 @@ def setup_exchanges(func) -> Callable[..., None]:
     """
     @wraps(func)
     def wrapper(*args, **kwargs):
-        with get_channel() as channel:
-            for exchange in DEFAULT_EXCHANGES:
-                channel.exchange_declare(exchange=exchange, exchange_type='fanout', durable=True)
-
+        declare_exchanges()
```
giorgiosironi, line 79: extraction is clear

docker-compose.yml
```diff
@@ -47,6 +47,16 @@ services:
...
         depends_on:
           - article-store-postgres
+    article-store-exchange-setup:
+        build:
+            context: ./article-store
+        volumes:
+            - ./article-store:/srv/app
+        environment:
+            - RABBITMQ_URL=amqp://guest:guest@event-bus:5672
+        command: pipenv run python manage.py setup_exchanges
```
giorgiosironi, line 57: doesn't block the other containers from progressing while these are declared, but they don't need to as they idempotently declare them as well just in case

giorgiosironi (approved):

PR #39, opened 2018-05-29
Update event format

thewilkybarkid:
To match #28.


article-store/articles/api.py
```diff
@@ -74,7 +75,7 @@ def post(self, request: Request, article_id: str) -> HttpResponse:
         with transaction.atomic():
             article, created = Article.objects.get_or_create(id=article_id)
 
-            with message_publisher('article.version.post', article_id, article.next_version):
+            with message_publisher('article.version.post', request.META.get(settings.RUN_ID_HEADER)):
```
giorgiosironi, line 78: `request` specific to the controller layer, `message_publisher` to the infrastructural layer around the models, that fits. Later some kind of creational pattern will emerge

article-store/core/middleware.py
```diff
@@ -0,0 +1,17 @@
...
+from django.conf import settings
+from django.http.request import HttpRequest
+from django.http.response import HttpResponse
+
+
+def ensure_run_id(get_response: Callable[[HttpRequest], HttpResponse]) -> Callable[[HttpRequest], HttpResponse]:
+    def middleware(request: HttpRequest):
+        run_id = request.META.get(settings.RUN_ID_HEADER, str(uuid.uuid1()))
+
+        request.META[settings.RUN_ID_HEADER] = run_id
```
giorgiosironi, line 13: a bit tied to Django, but ensure `runId` is consistent across the request handling

article-store/core/middleware.py
```diff
@@ -0,0 +1,17 @@
+from typing import Callable
+import uuid
+
+from django.conf import settings
+from django.http.request import HttpRequest
+from django.http.response import HttpResponse
+
+
+def ensure_run_id(get_response: Callable[[HttpRequest], HttpResponse]) -> Callable[[HttpRequest], HttpResponse]:
```
seanwiseman, line 9: Nice middleware solution 👍

article-store/core/settings.py
```diff
@@ -49,6 +49,7 @@
 ]
 
 MIDDLEWARE = [
+    'core.middleware.ensure_run_id',
```
giorgiosironi, line 52: unrelated, but the rest of the middlewares will need some cleanup

dashboard/src/Entity/Event.php
```diff
@@ -8,15 +8,10 @@
 
 /**
  * @ORM\Entity
- * @ORM\Table
+ * @ORM\Table(uniqueConstraints={@ORM\UniqueConstraint(columns={"run", "type"})})
```
giorgiosironi: this seems too strong a constraint now, for example preventing short retries of activities for temporary failures

giorgiosironi: de-duplication should be dealt with by the `@Id` field

giorgiosironi (commented):
Only point sticking out is the `UniqueConstraint`

dashboard/src/Entity/Event.php
thewilkybarkid: It'll probably become a constraint on `run`, `type`, and `attempt`. But let's ignore for now (since this is already simpler than we know it needs to be).

seanwiseman (approved):
Keen to get this in as soon as all parties are happy with it as it relates to the article store + airflow work I am in the process of adding.

PR #42, opened 2018-05-30
Article store messages

seanwiseman:
A small tidy up from messaging changes.

### DIFF
--- a/article-store/events/tests/test_event_utils.py
+++ b/article-store/events/tests/test_event_utils.py
@@ -9,15 +9,14 @@ from events.utils import (
 
 
 def test_can_get_base_message():
-    base_msg = create_message(msg_type='foo.test', identifier='someId', data={})
+    run_id = '2753d3ee-63de-11e8-9add-0242ac130008'
+    base_msg = create_message(msg_type='foo.test', run_id=run_id)
+
     assert base_msg
     assert len(base_msg['eventId'])
     assert base_msg['happenedAt']
     assert base_msg['type'] == 'foo.test'
-    assert base_msg['data'] == {}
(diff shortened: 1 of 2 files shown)


article-store/events/utils.py
```diff
@@ -80,7 +80,7 @@ def wrapper(*args, **kwargs):
 
 
 @setup_exchanges
-def send_article_message(msg_type: str, run_id: str, message: Optional[str]) -> None:
+def send_article_message(msg_type: str, run_id: str, message: Optional[str] = '') -> None:
```
thewilkybarkid: `None` rather than an empty string?

article-store/events/tests/test_event_utils.py
```diff
@@ -9,15 +9,14 @@
...
+
     assert base_msg
     assert len(base_msg['eventId'])
     assert base_msg['happenedAt']
     assert base_msg['type'] == 'foo.test'
-    assert base_msg['data'] == {}
-    assert base_msg['aggregate']['service'] == 'article-store'
-    assert base_msg['aggregate']['name'] == 'article-version'
-    assert base_msg['aggregate']['identifier'] == 'someId'
+    assert base_msg['runId'] == run_id
```
thewilkybarkid, line 19: Yeah, haven't been bothering with tests anywhere. #movefast #breakthings

seanwiseman: I have not been overdoing them, but having some present but failing was worth the fix.

article-store/events/utils.py
seanwiseman: Could argue it both ways, it's not being manipulated in the function so it `None` would still work.

thewilkybarkid (approved):

Issue #33, opened 2018-05-23
Article store Airflow integration


seanwiseman:
Add functionality to communicate with an Airflow instance in the form of one or more DAG(s) which demonstrate a logical and extensible integration.

- [x] Add Airflow services to project level `docker-compose.yml`
- [x] Draft requirements for tasks and a target DAG
- [x] Create tasks and DAG
- [x] Create basic adapter for the article store to talk to Airflow
- [x] Implement initiating the DAG from the article store

giorgiosironi:
With the current `master` I get these on `docker-compose up -V`:
```
article-store-airflow-webserver_1  | Fri May 25 15:37:41 UTC 2018 - waiting for Postgres... 18/20
article-store-airflow-flower_1     | Fri May 25 15:37:41 UTC 2018 - waiting for Redis... 19/20
article-store-airflow-scheduler_1  | Fri May 25 15:37:42 UTC 2018 - waiting for Postgres... 18/20
article-store-airflow-worker_1     | Fri May 25 15:37:42 UTC 2018 - waiting for Postgres... 18/20
article-store-airflow-webserver_1  | Fri May 25 15:37:46 UTC 2018 - waiting for Postgres... 19/20
article-store-airflow-flower_1     | Fri May 25 15:37:46 UTC 2018 - localhost:6379 still not reachable, giving up
walking-skeleton_article-store-airflow-flower_1 exited with code 1
article-store-airflow-scheduler_1  | Fri May 25 15:37:47 UTC 2018 - waiting for Postgres... 19/20
article-store-airflow-worker_1     | Fri May 25 15:37:48 UTC 2018 - waiting for Postgres... 19/20
article-store-airflow-webserver_1  | Fri May 25 15:37:51 UTC 2018 - localhost:5432 still not reachable, giving up
walking-skeleton_article-store-airflow-webserver_1 exited with code 1
article-store-airflow-scheduler_1  | Fri May 25 15:37:52 UTC 2018 - localhost:5432 still not reachable, giving up
walking-skeleton_article-store-airflow-scheduler_1 exited with code 1
article-store-airflow-worker_1     | Fri May 25 15:37:53 UTC 2018 - localhost:5432 still not reachable, giving up
```
The related postgres container seems alive and well, are these airflow containers misconfigured?
```
walking-skeleton_article-store-postgres_1   docker-entrypoint.sh postgres   Up      5432/tcp
```
[...]

PR #41, opened 2018-05-29
Add a demo

thewilkybarkid:

PR #43, opened 2018-05-31
Article store airflow integration

seanwiseman:
Carrying out the work specified in [#33](https://github.com/libero/walking-skeleton/issues/33).

Implemented as follows:

article-store-airflow mmd

Reasons for some decisions made:

**Not packaging up the libero airflow plugin and dags:**
I attempted to create and use a python package containing the plugin and dags functionality. This caused complications related to: mono repo + containers + python wheel file + pipenv. It was working but was not worth the complexity it added. 

For now its split between the `dags` and `plugin` directories in two modules. Looking ahead, this will be fine to be packaged up (put on `pypi` etc) and imported as a dependency.

**Not sending the entire XML content as inital param values to trigger the dag:**
Inital ideas pointed to sending the entire article content along with the initial values to trigger the dag. This caused issues as Airflows REST API wants these params to be passed in as a string which when parsed as json, is valid. 

This meant lots of additional escaping and head aches and was a nightmare to debug. It was cleaner to just pass in some required values and let the task(s) talk back to the `article-store` directly instead of this rather limited initial params passing method.

Going forward I think it would be worth looking at adding small plugin in the form of an API view so that we can better control this inital param passing.


article-store/articles/api.py
```diff
@@ -68,14 +75,15 @@ def post(self, request: Request, article_id: str) -> HttpResponse:
         :param article_id: str
         :return: class: `HttpResponse`
         """
-
         if not article_id or not Article.id_is_valid(article_id):
             return Response({'error': 'Please provide a valid article id'}, status=status.HTTP_406_NOT_ACCEPTABLE)
```
thewilkybarkid: `406 Not Acceptable` relates to `Accept` headers. Invalid input should be `400 Bad Request`

docker-compose.yml
```diff
@@ -43,6 +43,9 @@ services:
             - ALLOWED_HOSTS=article-store,localhost
             - APP_ENV=dev
             - APP_SECRET=f6a87c2e031ecadc0d7eb442729435c8
+            - AIRFLOW_ACTIVE=true
+            - AIRFLOW_API_HOST=article-store-airflow-webserver
+            - AIRFLOW_API_PORT=8080
```
thewilkybarkid: Should go for `AIRFLOW_URL` to match Postgres and RabbitMQ.

seanwiseman:
I would suggest creating article [`00666`](https://github.com/libero/walking-skeleton/tree/master/api-dummy/data/articles/00666/1/en) as this has quite a few assets in it.
You should `POST` to http://localhost:8000/articles/00666/versions using [this payload](https://gist.github.com/seanwiseman/0bcc91f5d54b3baffb4941df23b58e1e) then you should see:
- Dashboard events: http://localhost:8082/
- Airflow tasks: http://localhost:8086/admin/airflow/graph?dag_id=make_assets_public&root=

Once complete you should be able to see:
- The contents asset uris http://localhost:8000/articles/00666/versions/latest/body referencing `localhost:8089` 
- The article on the journal http://localhost:8080/articles/00666

thewilkybarkid:
Going to merge #41, could you update it to use that article?

seanwiseman:
> Going to merge #41, could you update it to use that article?

Any article will run through it now, 00666 just had assets which actually exercised the tasks in the dag. If you use 1234 then it should still trigger everything.

thewilkybarkid:
Seems to work. The emitted events aren't quite how I imagined they'd be (eg I thought `article.post.completed` would be the last event), but I think it's ok for now (since the principals are working).

Thinking about Airflow and DAGs, could they be part of the article store and mounted/copied into the article store airflow? Imagining other backends using Libero Airflow, but would have their own set of DAGS.

seanwiseman:
> Thinking about Airflow and DAGs, could they be part of the article store and mounted/copied into the article store airflow? Imagining other backends using Libero Airflow, but would have their own set of DAGS.

For the DAGS sure, it just really depends on how much reuse we want to get out of common tasks, if any. 

Though for plugins you would want it to live in something seperate I think like a library you could import. For example the `EventEmittingPythonOperator` should be available to all DAG providers and therefore should not live in any one project.

I think once we go past the one DAG stage we should experiement with breaking out the domain specific DAGS and keeping common DAGS, tasks and plugins in a library.

thewilkybarkid:
> For the DAGS sure, it just really depends on how much reuse we want to get out of common tasks, if any.

I imagine the article store would have a set of common DAGs, then a search service might have a set of common DAGs etc. Not sure there would be commonality between these though.

> Though for plugins you would want it to live in something seperate I think like a library you could import. For example the `EventEmittingPythonOperator` should be available to all DAG providers and therefore should not live in any one project.

Definitely, there should be a base Libero Airflow project that gets used by any service needing to use workflows.

workflow/airflow_libero/dags/make_assets_public_dag.py
```diff
@@ -0,0 +1,238 @@
...
+_fetch_article_content.set_upstream(_store_article_data)
+_extract_asset_uris.set_upstream(_fetch_article_content)
+_download_assets.set_upstream(_extract_asset_uris)
+_update_asset_uris.set_upstream(_download_assets)
+_deposit_to_article_store.set_upstream(_update_asset_uris)
+
+
+# TODO deposit_assets
+
+# TODO update_asset_uris
```
thewilkybarkid: Don't look to be TODOs anymore?

workflow/airflow_libero/dags/make_assets_public_dag.py
```diff
@@ -0,0 +1,238 @@
...
+        kwargs['ti'].xcom_push('article_version', input_data.get('article_version'))
+        kwargs['ti'].xcom_push('article_version_id', input_data.get('article_version_id'))
+        return SUCCESS
+
+    return FAILURE
+
+
+def fetch_article_content(*args, **kwargs) -> bool:
+    article_version_id = kwargs['ti'].xcom_pull(task_ids=None, key='article_version_id')
+    url = f'http://article-store:8000/articles/api/v1/article-versions/{article_version_id}/'
```
thewilkybarkid: `article-store:8000` should be configuration.

workflow/airflow_libero/dags/make_assets_public_dag.py
```diff
@@ -0,0 +1,238 @@
...
+    article_id = kwargs['ti'].xcom_pull(task_ids=None, key='article_id')
+    article_version = kwargs['ti'].xcom_pull(task_ids=None, key='article_version')
+    article_content = kwargs['ti'].xcom_pull(task_ids=None, key='article_content')
+
+    url = f'http://article-store:8000/articles/{article_id}/versions/{article_version}'
+    payload = convert_to_xml(article_content)
+
+    headers = {
+        'X-LIBERO-RUN-ID': run_id,
+        'X-LIBERO-AIRFLOW': 'true',
```
thewilkybarkid, line 179: I hadn't thought about using the public API for this communication, simpler than having another mechanism, but would probably need something more to limit Airflow bypassing to Airflow.

workflow/airflow_libero/dags/make_assets_public_dag.py
seanwiseman: There are a few values that should be provided as configuration here. Was part of a refactor that hasn't happened.

Will provide them as environment variables in the container to match other settings.

seanwiseman:
Is there anything further that needs adding or amending on this?

thewilkybarkid (approved):
Still a few things to think about, but the basic principals works so :shipit:.

PR #49, opened 2018-06-13
Separate Airflow from the Article Store

thewilkybarkid:
Not strictly necessary, but to help my knowledge of Airflow I've implemented https://github.com/libero/walking-skeleton/pull/43#issuecomment-396188124.


docker-compose.yml
```diff
@@ -62,10 +62,12 @@ services:
         command: >
                 /bin/bash -c '
                     until echo > /dev/tcp/article-store-postgres/5432; do sleep 1; done
+                    until echo > /dev/tcp/event-bus/5672; do sleep 1; done
```
thewilkybarkid, line 65: Unrelated, but was seeing occasional connection errors.

docker-compose.yml
```diff
@@ -121,12 +123,11 @@ services:
         image: 'redis:3.2.7'
     article-store-airflow-scheduler:
         build:
-          context: ./workflow/airflow_libero
+          context: ./airflow
         depends_on:
             - article-store-airflow-webserver
         volumes:
-            - ./workflow/airflow_libero/dags:/usr/local/airflow/dags
-            - ./workflow/airflow_libero/plugins:/usr/local/airflow/plugins
```
seanwiseman, line 129: I was getting an import error:

`ImportError: cannot import name 'EventEmittingPythonOperator`

It's because the `plugins` volume is still needed. I believe only the scheduler and the worker need it.

thewilkybarkid: It should be in the container...

thewilkybarkid: Do you have an old `__pycache__` in the `dags` folder that's being mounted?

seanwiseman: There was no `__pycache__` in the mounted `dags` folder but after a `down -v` and a fresh build it's working for me 👍

seanwiseman (approved):

workflow/airflow_libero/README.md
```diff
@@ -1,110 +0,0 @@
-# Airflow Example Usage
```
giorgiosironi, line 1: README is removed but not added elsewhere, was it a pure duplicate?

thewilkybarkid: We need to update the main README generally. This was out of date though.

docker-compose.yml
```diff
@@ -121,12 +123,11 @@ services:
...
+            - ./article-store/airflow/dags:/usr/local/airflow/dags
```
giorgiosironi, line 130: it is probably already true that the `dags` module can't use libraries, configuration or other modules from the article store, as it is mounted as a single folder inside airflow. If this is true it should also be separated from the article store, as it is its own independent application and never runs in the same process as the article store other Python modules?

seanwiseman: This is true. This was the direction I was pointing towards with having the `dags` and `plugins` in a separate lib.

thewilkybarkid: Ah, I though they were mounted for development purposes. I think the existing plugin is part of the base Libero (so does need the changes made here to separate it), though it should definitely be possible to more.

Agree that the DAGs should be separate, conceptually part of the article store so that's where I've put them for the minute.

Interesting question about libraries, how would a DAG use a library? Presumably the library has to be part of Airflow, or can each DAG have its own dependencies?

seanwiseman: Firstly, I think the current location (in this PR) is fine for the skeleton 👍 

> Presumably the library has to be part of Airflow

Yes you just need to make sure its available from the runtime that your airflow deployment uses.

At the core of it airflow just imports all dag classes it finds from the files (or imports within files) in the `dags` folder into its python `global` object. So there is room for pythonic magic if we so choose to cast it ⚡️

From my investigations most users just dump everything in the same `dags` folder and be done with it, how boring 😉 .

thewilkybarkid: Hmm, beyond my Python knowledge. Ideally you should just be able to dump _everything_ in the `dags` (and maybe `plugins` folder). Imagine a DAG needs a different version of a library (eg requests) used by Airflow, or a DAG needs a different version of a library used by a different DAG. There be a way round that? (And how would you install the library for a DAG without adding it to Airflow's `Pipfile`?)

seanwiseman: As I understand it, a DAG is just a collection of tasks to run, each DAG cannot be treated as having separate/different dependencies to the airflow instance it's running on. With that in mind, you need to be mind full of what DAGS are being included in an airflow instance as you will have to provide the dependencies for all its.

If you needed to have different versions to something the airflow requires itself... 🤷‍♂️  I think airflows attitude is ... "don't do that, a DAG is not its own application".

I don't think this should hold up this PR.

thewilkybarkid: Ok, in that case we should rethink it as being something you can drop code into to being an application like the others (Article Store, Journal etc). So we wouldn't provide a 'Libero Airflow' application/container, instead have a library for Libero integration into Airflow, a library of generic article DAGs, then an application/container for an 'Article Store Airflow' (ie `Pipfile` containing Airflow, Airflow-Libero library and article DAGs library). Dependencies are like any other project, if there's a mismatch in versions they can't be used together.

If you want to add your own DAGs (or reuse someone else's), it's then your own project ((ie `Pipfile` containing Airflow, Airflow-Libero library, `0..n` DAG libraries, and code for your own DAGs and their dependencies).

> I don't think this should hold up this PR.

Agreed; can't easily replicate the above here.

seanwiseman: Exactly 👍

giorgiosironi: I'm all for isolating Airflow into a single service in general. If we need to customize its dependencies that's the point of entry :+1: 
Not sure how the operators communicate data back in general, but if we/someone/ really needs a custom Python or custom process, there is the option of configuring other operators depending on `bash` or even `docker`: https://airflow.incubator.apache.org/concepts.html#operators

seanwiseman: > Not sure how the operators communicate data back in general

The operators communicate data back via its `XCOM` system, in our case `article-store-airflow-redis`.

> there is the option of configuring other operators depending on bash or even docker

But would still be executed within the same Python environment. For example, the `BashOperator` is just using a `subprocess.Popen` under the hood. The operators are just wrappers around a function that (in our case)  a Celery worker executes. That worker process is still based on the same dependencies/image as the other airflow components.

If a some task(s) required wildly different dependencies (for whatever reason ) than the airflow instance then you could just outsource that to a microservice which a DAG could just communicate with.

giorgiosironi: I assumed `BashOperator` could run non-Python stuff (e.g. `wget https://example.com/file`) hence also another Python `/srv/builder/venv/bin/python ...`; but yes the choice of a separate microservice to orchestrate is the more robust.

seanwiseman: Yes, it opens up the use of standard bash'ness and it would make Python `inception` possible.  😎

PR #52, opened 2018-06-18
Cleanup

thewilkybarkid:


thewilkybarkid:
README still needs improving.

README.md
```diff
@@ -2,22 +2,82 @@
...
+Contains a simple Nginx API gateway.
+
+* http://localhost:8085/ proxies `api-doc`
+* http://localhost:8085/articles proxies `article-store`
+
+### `article-store`
+
+Can be viewed through the `api-gateway` at http://localhost:8085/articles.
+
+Contains an article store
```
seanwiseman: Perhaps mention Django here as other frameworks/technologies have been mentioned on the other services.

README.md
```diff
@@ -87,4 +87,6 @@ A dummy implementation of the API, containing sample XML content in multiple lan
 - `journal` does not convert all content to HTML
 - `dashboard` is missing some event name translations
 - `article-store` persists article versions before the workflows are run
+- `article-store` doesn't validate input
```
seanwiseman: Was by design. Do we have a strategy for accessing the schemas(s)? if so this can be implemented.

thewilkybarkid: It would be configured with a schema somehow (currently seems to take it as input). Wouldn't worry about it too much.

thewilkybarkid:
Probably mergeable now, though I'm sure the README/example script could be improved.

seanwiseman (approved):
Seems fine for now. We can probably tweak/improve the readme and example.sh after running through it a few times ourselves in prep for the demo.

README.md
```diff
@@ -1,23 +1,96 @@
...
 
-It contains:
+1. Creating an article
+2. Adding a new version
+3. Updating a version
+4. Deleting an article
 
 ## Applications
 
 These can be run using `docker-compose up -V`.
```
giorgiosironi: Since there are many services here, one more level in the hierarchy (e.g. API-oriented vs user interface or public-oriented vs back office) may help understanding what they are

thewilkybarkid:
Could add a diagram of how the services fit together.

static-assets-store/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM nginx:1.13
+
+RUN mkdir /assets && \
+    chmod 2777 /assets
```
giorgiosironi, line 4: `2777`? For the sticky group?

thewilkybarkid: `article-store-airflow` writes as its `airflow` user. This is a bit brute-force, feel free to change but don't see it as important.

README.md
seanwiseman: I like the diagram 👍 . One thing, Airflow executes its DAGs and tasks completely independently of the Article store and therefore publishes events directly to the event bus. In the current diagram, it looks like it could only communicate through the Article store but should probably show the direct event bus connection.

thewilkybarkid: It's in https://drive.google.com/drive/folders/1MUU8ewCUWN0lwq5i7NCbA9IGwEf2-oel if you want to tweak. Trying to find a balance between factual and not-complicated!

thewilkybarkid:
:shipit:?

static-assets-store/Dockerfile
giorgiosironi: 777 is fine, is the `2` that sticks out, but not important at the moment

giorgiosironi (approved):

PR #53, opened 2018-06-27
Add wait for services in example.sh

seanwiseman:
Resolves an issue getting a 502 if you run through the `example.sh` and hit enter before the `article-store` is ready.

If you have word from the `bash` gods on how to refactor this addition please go ahead, but it does currently solve the issue.

### DIFF
--- a/examples.sh
+++ b/examples.sh
@@ -13,7 +13,7 @@ function line {
 function wait_for_services {
     echo "Waiting for services..."
     while true; do
-        STATUS_CODE=$(curl -o /dev/null --silent --head --write-out '%{http_code}\n' 'http://localhost:8085/articles')
+        STATUS_CODE=$(curl -o /dev/null --silent --head --write-out '%{http_code}' 'http://localhost:8085/articles')
         if [ "$STATUS_CODE" = "200" ]
         then
             break
--- a/examples.sh
+++ b/examples.sh
@@ -11,8 +11,10 @@ function line {
 }
 
 function wait_for_services {
+    TIMEOUT="${2:-30}"
     echo "Waiting for services..."
-    while true; do
+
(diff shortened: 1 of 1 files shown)


thewilkybarkid (approved):

examples.sh
```diff
@@ -10,10 +10,23 @@ function line {
     echo -e "\n"
 }
 
+function wait_for_services {
+    echo "Waiting for services..."
+    while true; do
+        STATUS_CODE=$(curl -o /dev/null --silent --head --write-out '%{http_code}\n' 'http://localhost:8085/articles')
```
giorgiosironi: don't need the `\n` I think

examples.sh
```diff
@@ -10,10 +10,23 @@ function line {
...
+        if [ "$STATUS_CODE" = "200" ]
+        then
+            break
+        fi
```
giorgiosironi: there should be a timeout because this could sleep indefinitely

giorgiosironi: `timeout` accepts an inline script, but it is error-prone to use it; if the function can be made an external script the filename itself can be passed (example of the error-prone inline script: https://github.com/elifesciences/builder-base-formula/blob/89a9de925e66593b378f6e03b701e89235fa0850/elife/docker-scripts/docker-wait-healthy#L14-L24)

seanwiseman: I tried adding it using `timeout` but it did not seem to like using the inline script method from within a `function`. You could perhaps just refactor the `start` `function` to be split and insert the `timeout` stuff in between, but I swapped out the `while` for a `for` with a limit. Will this suffice for the time being?

giorgiosironi (approved):
"for with limit" keeps the intent of avoiding an infinite wait

PR #54, opened 2018-06-29
Update example images

thewilkybarkid:
Puts an image in every version (and shows a change).


article-store-airflow/dags/make_assets_public_dag.py
```diff
@@ -135,6 +136,8 @@ def download_assets(*args, **kwargs) -> bool:
 
         for uri in uris['asset_uris']:
             response = requests.get(url=uri)
+            response.raise_for_status()
```
thewilkybarkid, line 139: I forgot to do a `docker-compose build`, so the new image didn't exist and resulted in a blank file being created rather than a failure.

seanwiseman (approved):

PR #57, opened 2018-07-26
[WIP] Experiment with moving code/templating from Journal to libraries

thewilkybarkid:
(This isn't meant to be merged, and is probably easier to read the end result rather than a diff. Is also incomplete.)

When doing the Journal here we didn't really worry about the front-end patterns and how to reuse code. We do need to look at this, so this is an experiment to see how it could work.

Assumes:

- We're using Twig for the patterns (agreed with @davidcmoulton )
- We want to have as little code in Journal as possible, so that creating your own applications means you could reuse code without having to fork it
- We want to allow extending/replacing patterns, without having to replace/duplicate the whole lot.


journal/composer.json
```diff
@@ -31,10 +32,13 @@
     },
     "autoload": {
         "psr-4": {
-            "Libero\\Journal\\": "src/"
+            "Libero\\Journal\\": "src/",
+            "Libero\\": "vendor-extra/"
```
thewilkybarkid, line 36: `vendor-extra` contains sample extracted libraries.

journal/src/Controller/ArticleController.php
```diff
@@ -2,94 +2,82 @@
...
 use Symfony\Component\HttpFoundation\Response;
 use Twig\Environment;
-use function extract;
+use function Functional\const_function;
+use function Functional\map;
 use function GuzzleHttp\Promise\all;
-use function Libero\Journal\will_convert_all;
+use function Libero\PatternsBundle\locale_direction;
 
 final class ArticleController
```
thewilkybarkid, line 17: This could be moved to a library. Not yet looked into how it could be extended.

journal/src/Controller/HomeController.php
```diff
@@ -2,61 +2,74 @@
...
 final class HomeController
 {
+    private $articlesClient;
+    private $converter;
+    private $translator;
     private $twig;
-    private $apiClient;
 
-    public function __construct(Environment $twig, ClientInterface $apiClient)
+    public function __construct(ArticlesClient $articlesClient, Environment $twig, ViewConverter $converter, TranslatorInterface $translator)
```
thewilkybarkid, line 26: Same as https://github.com/libero/walking-skeleton/pull/57/files#r205458058.

journal/src/View/Converter/Block/AsideConverter.php
```diff
@@ -1,137 +0,0 @@
-<?php
-
-namespace Libero\Journal\View\Converter\Block;
-
-use Libero\Journal\Dom\Element;
-use Libero\Journal\View\Converter\ViewConverter;
-use Symfony\Component\Translation\TranslatorInterface;
-use const Libero\Journal\LIBERO;
-
-final class AsideConverter implements ViewConverter
```
thewilkybarkid, line 10: Not all of these have been replaced.

journal/src/Views/ViewConverter/Blocks/eLifeFrontContentHeaderConverter.php
```diff
@@ -0,0 +1,81 @@
...
+        if (!$object instanceof Element || '@LiberoPatterns/patterns/content-header.html.twig' !== $template || 'front' !== $object->getName() || LIBERO !== $object->getNamespace()) {
+            return null;
+        }
+
+        $context['level'] = ($context['level'] ?? 1) + 1;
+
+        return new View(
+            $template, array_filter(
+                [
+                    'digest' => $this->digest($object, $context),
```
thewilkybarkid, line 39: This is an example of extending the Libero content header pattern with custom data.

journal/src/Views/ViewConverter/Inline/ColorConverter.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public function convert(Node $object, ?string $template, array $context = []) : ?View
+    {
+        if (!$object instanceof Element || 'color' !== $object->getName() || ELIFE !== $object->getNamespace()) {
+            return null;
+        }
+
+        return new View(
+            '@LiberoPatterns/patterns/span.html.twig', [
```
thewilkybarkid, line 31: This is an example of adding a custom inline element. Reuses a generic Libero template, but could use its own.

journal/templates/bundles/LiberoPatternsBundle/patterns/content-header.html.twig
```diff
@@ -0,0 +1,11 @@
+{% extends '@!LiberoPatterns/patterns/content-header.html.twig' %}
```
thewilkybarkid, line 1: This is an example of replace a single template, then _extending_ it. So the whole thing isn't replaced. (Though could be if you want.)

journal/translations/messages.ar.yaml
```diff
@@ -13,4 +13,3 @@ app:
     language_choice: 'اختار اللغة:'
     no_articles: 'لا توجد مقالات متاحة.'
     title: 'عرض Libero'
-direction: rtl
```
thewilkybarkid, line 16: This was always a temporary measure, replaced with Punic (which contains CLDR data).

journal/vendor-extra/ApiClient/ArticlesClient.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+namespace Libero\ApiClient;
+
+use GuzzleHttp\Promise\PromiseInterface;
+use GuzzleHttp\Psr7\Request;
+
+final class ArticlesClient
```
thewilkybarkid, line 8: Provides a client for the defined Libero articles API.

If you choose to extend the API spec, like https://github.com/libero/walking-skeleton/blob/e89141108b1f3a8208d852ebe5aeed801a730681/api-doc/api.raml#L20-L59 you'll need your own client. (_Might_ be something we could do about that?)

journal/vendor-extra/PatternsBundle/Resources/views/patterns/content-header.html.twig
```diff
@@ -0,0 +1,13 @@
+<div{% include '@LiberoPatterns/attributes.html.twig' %}>
+
+    {% block content %}
```
thewilkybarkid, line 3: Example of allowing extension/replacement.

journal/vendor-extra/PatternsBundle/Resources/views/patterns/math.html.twig
```diff
@@ -0,0 +1 @@
+<span>{{ math|raw }}</span>{{ '' -}}
```
thewilkybarkid, line 1: `{{ '' -}}` is just to stop unnecessary whitespace (without removing the newline at the end of the file).

journal/vendor-extra/Views/ViewConverter.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Libero\Views;
+
+use Libero\Dom\Node;
+
+interface ViewConverter
+{
+    public function convert(Node $object, ?string $template, array $context = []) : ?View;
```
thewilkybarkid, line 9: Better to use `Element` rather than `Node`, but means handling text nodes separately everywhere.

journal/vendor-extra/Views/ViewConverter/StopwatchViewConverter.php
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Libero\Views\ViewConverter;
+
+use Libero\Dom\Element;
+use Libero\Dom\Node;
+use Libero\Views\View;
+use Libero\Views\ViewConverter;
+use Symfony\Component\Stopwatch\Stopwatch;
+
+final class StopwatchViewConverter implements ViewConverter
```
thewilkybarkid, line 11: So details appear in the Symfony Profiler performance tab.

journal/vendor-extra/Views/ViewConverter/TextConverter.php
```diff
@@ -0,0 +1,20 @@
...
+
+final class TextConverter implements ViewConverter
+{
+    public function convert(Node $object, ?string $template, array $context = []) : ?View
+    {
+        if (!$object instanceof Text) {
+            return null;
+        }
+
+        return new View('@LiberoPatterns/patterns/text.html.twig', ['text' => $object->toText()]);
```
thewilkybarkid, line 18: Clearly a bit weird to render plain text with a template. See https://github.com/libero/walking-skeleton/pull/57/files#r205463181.

journal/vendor-extra/Views/ViewConverter/ViewConverterRegistry.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $view = null;
+
+        foreach ($this->registry as $converter) {
+            $converted = $converter->convert($object, $template, $context);
+
+            if (!$view instanceof View) {
+                $view = $converted;
+            } elseif ($converted instanceof View) {
+                $view = $view->withArguments($converted->getArguments())->withTemplate($converted->getTemplate());
```
thewilkybarkid, line 28: This allows the results of `eLifeFrontContentHeaderConverter` and `FrontContentHeaderConverter` to be combined.

Might want to think about prioritisation, to allow changing of arguments. (Currently ambitious as to what order they run in.)

journal/vendor-extra/Views/ViewConverter/ViewConverterRegistry.php
```diff
@@ -0,0 +1,38 @@
...
+            }
+        }
+
+        if (null === $view && null === $template) {
+            return new View('@LiberoPatterns/patterns/text.html.twig', ['text' => "{$object->toText()}"]);
```
thewilkybarkid, line 33: So just displays the text of anything it doesn't know about (ie there's no converter). (No `$template` assuming it's an inline element.)

journal/vendor-extra/Views/ViewConverter/ViewConverterRegistry.php
thewilkybarkid: Could maybe pass an empty `View` instead of `$template` in `convert()` and let them populate it as they go along.

journal/vendor-extra/PatternsBundle/Resources/views/attributes.html.twig
```diff
@@ -0,0 +1,9 @@
+{% if attributes is defined %}
```
thewilkybarkid, line 1: These templates should be in a separate library (to represent the product of the pattern library), but for loading simplicity I've kept them inside the bundle.

docker-compose.yml
```diff
@@ -40,11 +40,11 @@ services:
             - ./api-gateway/.docker/nginx-server.conf.template:/etc/nginx/conf.d/default.conf.template
         command: sh -c "envsubst \"`env | awk -F = '{printf \" $$%s\", $$1}'`\" < /etc/nginx/conf.d/default.conf.template > /etc/nginx/conf.d/default.conf && nginx -g 'daemon off;'"
         environment:
-            - ARTICLES=http://article-store:8000
+            - ARTICLES=http://api-dummy-nginx:80
```
thewilkybarkid, line 43: Easier to see examples.

journal/vendor-extra/PatternsBundle/Resources/views/views.html.twig
```diff
@@ -0,0 +1,10 @@
+{% for view in views -%}
+
+    {% if view.template is defined %}
+        {%- include view.template with view.arguments only -%}
```
thewilkybarkid, line 4: This allows embedding arbitrary templates. Avoids the annoyance with eLife's Mustache templates where you have to render things early then include them as strings.

journal/vendor-extra/PatternsBundle/Resources/views/views.html.twig
```diff
@@ -0,0 +1,10 @@
...
+    {% else %}
+        {#- Assuming plain string -#}
+        {{- view -}}
```
thewilkybarkid, line 7: Not sure about this. Added for https://github.com/libero/walking-skeleton/blob/c28487fb3d36115eb0d4996d5c4680edbab0b108/journal/src/Controller/HomeController.php#L57 so it didn't have to manually create `text` views.

journal/vendor-extra/Dom/Document.php
```diff
@@ -15,27 +16,53 @@
...
     use FindsThroughXPath;
     use ReadOnlyArrayAccess;
+    use XmlAsString;
 
     private $document;
     private $namespaces = [];
     private $xPath;
 
-    public function __construct(string $xml, string $defaultNamespace)
+    public function __construct(Stopwatch $stopwatch, string $xml, string $defaultNamespace)
```
thewilkybarkid, line 25: Added `Stopwatch` so details appear in the Symfony Profiler performance tab.

journal/src/Views/ViewConverter/Blocks/eLifeFrontContentHeaderConverter.php
giorgiosironi: Uhm no references to the default `Libero\PatternsBundle\Views\ViewConverter\Blocks\FrontContentHeaderConverter`, how does this data get mixed with the fields produced there?

journal/src/Views/ViewConverter/Blocks/eLifeFrontContentHeaderConverter.php
```diff
@@ -0,0 +1,81 @@
...
+
+    private function digest(Element $object, array $context = []) : ?array
+    {
+        $digestElement = $object->get('elife:digest');
+
+        if (!$digestElement instanceof Element) {
+            return null;
+        }
+
+        return [
```
giorgiosironi, line 53: this looks like a generic view model object, maybe an extension to `ArrayObject` could help it be both array-like and add some validation (for standard fields like `attributes`) or behavior (utility methods that suddenly have a place to live?)

journal/vendor-extra/Views/ViewConverter/StopwatchViewConverter.php
giorgiosironi: The direction of the dependency looks like it could be reversed, e.g. a `ViewConverter` emits events and there is a listener which could start/stop the stopwatch

journal/src/Views/ViewConverter/Blocks/eLifeFrontContentHeaderConverter.php
thewilkybarkid: They get combined: https://github.com/libero/walking-skeleton/blob/c28487fb3d36115eb0d4996d5c4680edbab0b108/journal/vendor-extra/Views/ViewConverter/ViewConverterRegistry.php#L22-L30

journal/vendor-extra/Views/ViewConverter.php
thewilkybarkid: Now thinking this should be split up: have separate block (which can just use `Element`s, and require a template), and inline (which either use text as well or handle them different, and have no set template) converters.

journal/vendor-extra/Views/ViewConverter/ViewConverterRegistry.php
thewilkybarkid: Related to https://github.com/libero/walking-skeleton/pull/57#discussion_r208680885, these should be separated. So delegating the actual conversion to visitors, that take a `View` and add details as necessary.

PR #58, opened 2018-07-27
Add basic docker-compose build, Travis CI version

giorgiosironi:
A test of performance for Travis CI.


.travis.yml
```diff
@@ -0,0 +1,13 @@
+sudo: required
+
+# not just that, but we need to pick one
+language: python
```
thewilkybarkid: There is `minimal`, which is equivalent to `bash` (IIRC).

giorgiosironi:
Circle CI has some problems with mounting volumes in `docker-compose`, because of its use of remote docker. That is a showstopper unless solved, may not be ready for prime time.

giorgiosironi:
Disabled the Circle CI builds.

seanwiseman (approved):

